PR #32, opened 2022-03-01
feat: add logout api route

thorwebdev:
- Adds the `/api/auth/logout` route: Your Next.js application logs out the user. (you can optionally pass a `returnTo` parameter to return to a custom relative URL after logout, eg `/api/auth/logout?returnTo=/login`).

This flow is preferred as it works better with the `withAuthRequired` SSR approach.


kiwicopple:
Nice @thorwebdev - this route is very useful. Perhaps we could provide a way to overrde the redirect at instantiation?

Something like:

```js
export default handleAuth({
  logout: { returnTo: '/login' }
});
```

thorwebdev:
Yes, good idea, added ✅

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 1.2.0 - 2022-03-01
+
+- [BREAKING CHANGE][#32](https://github.com/supabase-community/supabase-auth-helpers/pull/32): feat: add logout api route. Note that this includes a breaking change to the options parameter for `handleAuth(options: HandleAuthOptions)` See [the docs](./src/nextjs/README.md#basic-setup) for more details.
```
silentworks, line 5: I don't think this is a breaking change since its a optional parameter with a default.

silentworks (approved):

CHANGELOG.md
thorwebdev: It is a breaking change if you were previously supplying the optional cookie options. They now need to be nested.

NicholasG04:
Just wondering in what way this is better than using `supabaseClient.auth.api.signOut()` please, since it requires redirection? Can we continue to use this, it seems to be working fine?

thorwebdev:
@NicholasG04 it is just a bit more convenient than the client side `signOut` as you don't need to manually redirect the user after signout.

E.g. when calling `supabaseClient.auth.signOut()` e.g. within a layout component, you then need to check which page they are on and if they are on a restricted page you need to redirect them away for example. So when you work a lot with SSR using `withAuthRequired` for example, it's more convenient to just redirect to a logout route.

NicholasG04:
Makes sense, thank you very much!

PR #42, opened 2022-03-16
fix: update cookie values on request object.

### LINKED ISSUE #41: Cache new tokens within same request
Currently, when you use `withAuthRequired` on your API route for example and `getUser` in your API route (e.g. https://github.com/vercel/nextjs-subscription-payments/blob/main/pages/api/create-portal-link.ts) and the access token is expired, the `getUser` call will throw an `Invalid Refresh Token` error because `withAuthRequired` has updated the tokens, but because we're in the same request, the cookies haven't been updated yet.

thorwebdev:
Fixes #41

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 1.2.3 - 2022-03-16
+
+- [#42](https://github.com/supabase-community/supabase-auth-helpers/pull/42): fix: update cookie values on request object.
+
 ## 1.2.2 - 2022-03-09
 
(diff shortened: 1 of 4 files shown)


alaister (commented):
Code looks good!

Something to think about is this functionality will not work with Next.js middleware, as the request object used in the middleware is different from the request object used elsewhere.

Do we just keep this functionality for normal Next.js and make users aware that using it with middleware is an edge case?

thorwebdev:
> Do we just keep this functionality for normal Next.js and make users aware that using it with middleware is an edge case?

Hm, that's a good point. Do we know why and how folks want to use this in middleware? Once we're clear about the use case, we can understand it better.

Also, maybe middleware does update the actually cookies before the next request hits? In that case it would be fine.

Issue #36, opened 2022-03-10
Improvement request: Compatibility with Next's middleware


Xananax:
Currently, `withAuthRequired` cannot be used in [middlewares](https://nextjs.org/docs/middleware). However, the difference between `getServerSideProps` and `middleware` is minimal.

By changing the function's requirements to be the strict minimum (access to cookies), I believe I've successfully made a Next middleware which can be shared with a `withAuthRequired` implementation, provided an additional small interface is created.

I don't feel that I grasp well enough how either Next or Supabase to do a proper PR, but here's my implementation:

```ts
// _middleware.ts

import type { NextFetchEvent, NextRequest } from "next/server";
import { NextResponse } from "next/server";
import { CookieSerializeOptions } from "next/dist/server/web/types";
import { GetServerSideProps, GetServerSidePropsContext } from "next";
import { jwtDecoder } from "@supabase/supabase-auth-helpers/shared/utils/jwt";
import { CookieOptions } from "@supabase/supabase-auth-helpers/nextjs/types";
import { COOKIE_OPTIONS } from "@supabase/supabase-auth-helpers/shared/utils/constants";
import { User, createClient, ApiError } from "@supabase/supabase-js";

type Cookies =
  | GetServerSidePropsContext["req"]["cookies"]
  | NextRequest["cookies"];

export async function getUser(
  cookies: Cookies,
  cookieAdditionalOptions?: CookieOptions
): Promise<{ user: User | null; accessToken: string | null; error?: string }> {
  try {
    if (
      !process.env.NEXT_PUBLIC_SUPABASE_URL ||
      !process.env.NEXT_PUBLIC_SUPABASE_ANON_KEY
    ) {
      throw new Error(
        "NEXT_PUBLIC_SUPABASE_URL and NEXT_PUBLIC_SUPABASE_ANON_KEY env variables are required!"
      );
    }
    if (!cookies) {
      throw new Error("Not able to parse cookies!");
    }
    const supabase = createClient(
      process.env.NEXT_PUBLIC_SUPABASE_URL,
      process.env.NEXT_PUBLIC_SUPABASE_ANON_KEY
    );
    const cookieOptions = { ...COOKIE_OPTIONS, ...cookieAdditionalOptions };
    const access_token = cookies[`${cookieOptions.name}-access-token`];
    const refresh_token = cookies[`${cookieOptions.name}-refresh-token`];

    if (!access_token) {
      throw new Error("No cookie found!");
    }
[...]

PR #44, opened 2022-03-16
feat: Add `authMiddleware` Nextjs Middleware util to protect directories

Saeris:
This adds a new utility function `authMiddleware` which leverages Next v12's [Middleware](https://nextjs.org/docs/middleware) feature to add blanket protection for all routes in a given directory via a `_middleware` file.

It is based on `getUser`, which will attempt to recover a user's session via their `access_token` or `refresh_token`. If a user's session is verified to exist, the incoming request is forwarded. Otherwise, they will be redirected to the given route, defaulting the the site root.

A usage example was also added to the README

### DIFF
--- a/src/nextjs/README.md
+++ b/src/nextjs/README.md
@@ -261,3 +261,14 @@ export default withAuthRequired(async function ProtectedRoute(req, res) {
 ```
 
 If you visit `/api/protected-route` without a valid session cookie, you will get a 401 response.
+
+## Protecting routes with [Nextjs Middleware](https://nextjs.org/docs/middleware)
+
+As an alternative to protecting individual routes using `getServerSideProps` with `withAuthRequired`, `authMiddleware` can be used from inside a `_middleware` file to protect an entire directory. In the following example, all requests to `/protected/*` will check whether a user is signed in, if successful the request will be forwarded to the destination route, otherwise the user will be redirected to `/login` (defaults to: `/`) with a 307 Temporary Redirect response status:
+
+```ts
+// pages/protected/_middleware.ts
+import { authMiddleware } from '@supabase/supabase-auth-helpers/nextjs';
+
+export const middleware = authMiddleware('/login');
+```
--- a/src/nextjs/README.md
+++ b/src/nextjs/README.md
@@ -193,52 +193,6 @@ export const getServerSideProps = withAuthRequired({
 });
 ```
 
-### Server-side data fetching to OAuth APIs using `provider_token`
-
-When using third-party auth providers, sessions are initiated with an additional `provider_token` field which is persisted as an HTTPOnly cookie upon logging in to enabled usage on the server side. The `provider_token` can be used to make API requests to the OAuth provider's API endpoints on behalf of the logged-in user. In the following example, we fetch the user's full profile from the third-party API during SSR using their id and auth token:
-
-```js
-import {
-  User,
-  withAuthRequired,
-  getUser
-} from '@supabase/supabase-auth-helpers/nextjs';
-
-interface Profile {
-  /* ... */
-}
-
-export default function ProtectedPage({
-  user,
-  data
-}: {
-  user: User,
-  profile: Profile
-}) {
-  return <div>Protected content</div>;
-}
-
-export const getServerSideProps = withAuthRequired({
-  redirectTo: '/',
-  async getServerSideProps(ctx) {
-    // Retrieve provider_token from cookies
-    const provider_token = ctx.req.cookies['sb-provider-token'];
-    // Get logged in user's third-party id from metadata
-    const { user } = await getUser(ctx);
-    const userId = user?.user_metadata.provider_id;
-    const profile: Profile = await (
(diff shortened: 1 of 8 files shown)


Saeris:
Fixes #36

src/nextjs/README.md
```diff
@@ -215,3 +215,14 @@ export default withAuthRequired(async function ProtectedRoute(req, res) {
...
+
+## Protecting routes with [Nextjs Middleware](https://nextjs.org/docs/middleware)
+
+As an alternative to protecting individual routes using `getServerSideProps` with `withAuthRequired`, `authMiddleware` can be used from inside a `_middleware` file to protect an entire directory. In the following example, all requests to `/protected/*` will check whether a user is signed in, if successful the request will be forwarded to the destination route, otherwise the user will be redirected to `/login` (defaults to: `/`) with a 307 Temporary Redirect response status:
+
+```ts
+// pages/protected/_middleware.ts
+import { authMiddleware } from '@supabase/supabase-auth-helpers/nextjs';
+
+export const middleware = authMiddleware('/login');
```
thorwebdev, line 227: Can you update to adhere to the following shape?

```js
export const middleware = withMiddlewareAuthRequired({ redirectTo: '/login' });
```

We're actually looking to break `withAuthRequired` out into `withPageAuthRequired` and `withApiAuthRequired `

src/nextjs/utils/authMiddleware.ts
```diff
@@ -0,0 +1,89 @@
...
+    } catch (err: unknown) {
+      if (err instanceof Error) {
+        console.log(
+          `Could not authenticate request, redirecting to ${redirectTo}:`,
+          err
+        );
+      }
+      const url = req.nextUrl.clone();
+      url.pathname = redirectTo;
+      url.searchParams.set(`ret`, req.nextUrl.pathname);
```
thorwebdev, line 85: You okay to call this param `nextUrl` instead?

thorwebdev (changes requested):
Thanks so much for this. Couple of nits I'd like to get your thoughts on.

src/nextjs/README.md
Saeris: Sure, sounds good to me!

src/nextjs/utils/authMiddleware.ts
Saeris: No problem, name here is arbitrary. Perhaps `redirectUrl` would be more descriptive?

Because we're constructing a new URL by taking the base of the existing nextUrl (ie: `https://localhost:3000` or `https://example.com`) and replacing it's path (`/dashboard` with `/login`), appending as a query parameter the path the user is bounced from.

thorwebdev: Yah, I like `redirectUrl` 👍

thorwebdev: Or could be `redirectTo` also I guess.

src/nextjs/utils/withMiddlewareAuthRequired.ts
```diff
@@ -0,0 +1,91 @@
...
+    } catch (err: unknown) {
+      if (err instanceof Error) {
+        console.log(
+          `Could not authenticate request, redirecting to ${redirectTo}:`,
+          err
+        );
+      }
+      const redirectUrl = req.nextUrl.clone();
+      redirectUrl.pathname = redirectTo;
+      redirectUrl.searchParams.set(`ret`, req.nextUrl.pathname);
```
alaister, line 87: Thanks for these changes!

Looks like the URL param is still called `ret` instead of `redirectUrl`. Would you mind updating it again?

alaister (changes requested):

src/nextjs/utils/withMiddlewareAuthRequired.ts
Saeris: ah, sorry I misunderstood. I thought you wanted the variable name changed. How about for the param name, we go with `redirectedFrom`?

thorwebdev: Yes, I think that works, thanks 👍

thorwebdev (approved):
LGTM, thanks 💚

thorwebdev:
@Saeris did you run into this dual esm/cjs issue at all? And if so, any recommendation how to resolve. For now I changed `import { NextResponse } from 'next/server';` to `import { NextResponse } from 'next/dist/server/web/spec-extension/response';` but that doesn't seem too ideal...

Saeris:
> did you run into this dual esm/cjs issue at all?

Hmm, not on my end. I don't know if this has anything to do with it, but I did notice that ya'll aren't using the recommended type only imports (`import type`) syntax that was introduced in [Typescript 3.8](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html)

Personally, I have my eslint rules setup such that these imports are properly handled for me as an autofix codemod.

Just venturing a guess here that by using this syntax, Typescript will resolve the type from a different file. Not sure what else to do beyond that.

PR #45, opened 2022-03-19
chore: clean up

thorwebdev:
- Adds provider_token example
- Adds middleware example
- Tries to make middleware token refreshing work correctly (it does seem to work locally, but not sure if it'll work correctly on the edge)...


eetzkorn:
Is it possible to add a provider_token client side example? I can't figure that out. Is the provider_token only available server_side with this package?

PR #48, opened 2022-03-24
don't catch errors in user handlers

a-type:
Moves code which invokes logic defined by the library user outside of the try-catch blocks, so that errors in user-written code are handled (or not) as normally expected by Next.

I'm a little unfamiliar with Next middleware so I'm not confident my moving of `NextResponse.next` was quite correct, let me know.

### DIFF
--- a/src/nextjs/utils/withAuthRequired.ts
+++ b/src/nextjs/utils/withAuthRequired.ts
@@ -91,15 +91,15 @@ export default function withAuthRequired(
           { cookieOptions }
         );
         if (!accessToken) throw new Error('No access token!');
-        await arg(req, res);
       } catch (error) {
         res.status(401).json({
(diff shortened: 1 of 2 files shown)


thorwebdev:
@a-type terribly sorry, this slipped through the cracks in the lead-up to Launch Week 🙈 I was wondering, would you be up for also doing https://github.com/supabase-community/supabase-auth-helpers/issues/59 within this PR?

a-type:
I'd prefer to leave it as-is, honestly - I didn't end up using Supabase as an auth provider, just wanted to contribute toward one pain point I had in trying it out, but I've had to move on to other work.

thorwebdev:
Adressed in https://github.com/supabase-community/supabase-auth-helpers/pull/92

PR #67, opened 2022-04-10
onAuthStateChangeFn, cleanup

jordiup:
Grants you access to the onAuthStateChangeFn, allows developers to hook into the callback


thorwebdev:
Thanks @jordiup, but there shouldn't really be a need to listen to the event yourself. Can you outline how you're using it?

PR #70, opened 2022-04-14
fix: better build pipeline

### LINKED ISSUE #22: Better release pipeline
Related to #14 

Currently we build to `/dist` folder and then run `semantic-release` within the `/dist` folder. However this means we have to have a separate `package.json` in the `/src` folder.

In the future a setup via separate packages and packaging and tree shaking via rollup might be better.

thorwebdev:
Fixes #22 

Okay, so the new packages setup works in the example, but I have no idea if the rollup config is correct, and how to test that.

And, unfortunately I haven't found a way to share code across packages, unless the package itself is being built and published to npm, maybe as `@supabase/auth-helpers-shared`?

Anyway, would appreciate you having a look and especially looking at the rollup stuff.


thorwebdev:
Closing in favour of https://github.com/supabase-community/supabase-auth-helpers/pull/75 👍

PR #75, opened 2022-04-18
Change repo to a monorepo using Turborepo

silentworks:
Add tsup as the bundler for all packages
Using pnpm as package manager instead of npm


README.md
```diff
@@ -11,3 +11,44 @@ A collection of framework specific Auth utilities for working with Supabase.
 
 - [Remix](https://github.com/supabase-community/supabase-auth-helpers/issues/57)
 - [SvelteKit](https://github.com/supabase-community/supabase-auth-helpers/issues/54)
+
+### Apps and Packages
+
+- `nextjs`: a [Next.js](https://nextjs.org) app
```
thorwebdev, line 17: Is this just used to define the term of `nextjs`?

README.md
```diff
@@ -11,3 +11,44 @@ A collection of framework specific Auth utilities for working with Supabase.
...
+
+This turborepo has some additional tools already setup for you:
+
+- [TypeScript](https://www.typescriptlang.org/) for static type checking
+- [ESLint](https://eslint.org/) for code linting
+- [Prettier](https://prettier.io) for code formatting
+
+## Setup
+
+This repository is used in the `npx create-turbo@latest` command, and selected when choosing which package manager you wish to use with your monorepo (PNPM).
```
thorwebdev, line 36: I think this step isn't entirely clear. Do I need to run `npx create-turbo@latest`? Can you make this clearer? how to I install the necessary dependencies?

thorwebdev: Or are you saying that this repo was set up via `npx create-turbo@latest`? But then do I just have to run pnpm install?

packages/nextjs/src/handlers/logout.ts
```diff
@@ -1,12 +1,12 @@
-import { CookieOptions } from '../types';
+import { CookieOptions } from 'shared/types';
 import type { NextApiRequest, NextApiResponse } from 'next';
-import { setCookies } from '../../shared/utils/cookies';
+import { setCookies } from 'shared/utils/cookies';
```
thorwebdev, line 3: One thing I noticed is that the code imported from `shared` is duplicated in each file that has an import for it. e.g. search for `// ../shared/` in the `nextjs/dist` folder. 

Is there a better way of doing this? E.g. could we tell the bundler to put the shared code into it's own files and import in the other files?

packages/react/src/components/UserProvider.tsx
```diff
@@ -31,6 +30,7 @@ export interface Props {
   profileUrl?: string;
   user?: User;
   fetcher?: UserFetcher;
+  pathname: any;
```
thorwebdev, line 33:
```suggestion
  pathname: string;
```

should this be `string`?

packages/react/package.json
```diff
@@ -0,0 +1,43 @@
+  {
+  "name": "@supabase/auth-helpers-react",
```
thorwebdev, line 2: Also need to change the neame in `packages/react/README.md`

package.json
```diff
@@ -1,65 +1,21 @@
...
-  "description": "A collection of framework specific Auth utilities for working with Supabase.",
-  "engines": {
-    "node": "^10.13.0 || >=12.0.0"
-  },
+  "private": true,
   "scripts": {
-    "clean": "rimraf dist",
-    "pretty": "prettier --write \"src/**/*.{ts,tsx}\" \"src/*.ts\"",
-    "lint": "tsc --noEmit && eslint --fix --ext .ts ./src",
-    "docs": "typedoc --options typedoc.js src",
```
thorwebdev, line 12: We should be able to keep the typedoc stuff and just change the command to `"docs": "typedoc --options typedoc.js packages"`, no?

thorwebdev (changes requested):
@silentworks this is a work of beauty! 😍 

Just some questions and nits. And I think we'll have to change the `.github/workflows/release.yml` action, right?

README.md
silentworks: I should remove that line because its saying the repository was setup with that command. Yes you would just need to install pnpm and run pnpm install at the root level to get everything installed.

packages/nextjs/src/handlers/logout.ts
silentworks: Good question, I'll have a look into this as it will be duplicate code in each file otherwise.

README.md
silentworks: No, it's used to tell what kind of apps are in the `examples` directory and the `packages` directory. Maybe we should give these actual names.

package.json
silentworks: Is this on a per package basis? where will this be used?

packages/react/package.json
silentworks: Yes for sure

packages/react/src/components/UserProvider.tsx
silentworks: Yeah I think you are correct here.

packages/nextjs/src/handlers/logout.ts
silentworks: Ok so the issue here is that because I'm bundling the code it's pulling in all dependencies into the one file. If we truly want to go the bundling route then we would only bundle the intex.ts file since it contains the import/export of all modules. If we want to go the non-bundling route which is what we've been doing in the past then we need to make the `shared` package an actual `npm package`, so we would publish it under `@supabase/auth-helpers-shared-lib` and make it a dependency to `@supabase/auth-helpers-nextjs` (`@supabase/auth-helpers-sveltekit` when its ready.)

README.md
thorwebdev: Then maybe call it "Packages & Examples" and make a nested list?

- Packages
  - ...
- Examples
  - [Next.js](./examples/nextjs)

README.md
thorwebdev: Okay, yes, then let's remove it and just say

- Install [pnpm](https://pnpm.io/installation)
- Run `pnpm install`

or something along those lines

package.json
thorwebdev: It runs on merge (https://github.com/supabase-community/supabase-auth-helpers/blob/next/.github/workflows/docs.yml) to publish https://supabase-community.github.io/supabase-auth-helpers/

packages/nextjs/src/handlers/logout.ts
thorwebdev: Okay, I think the bundling is fine for now. The shared code is very light so far.

package.json
silentworks: Ok this is a bit more tricky to get working in a monorepo. I've seen a few examples using it but I'm still trying to figure out how to get it working in our setup.

thorwebdev (approved):
LGTM 🥳

tinezmatias:
When are you going to make a release with these changes?

Issue #73, opened 2022-04-17
The users JWT expires if on same page longer than expire time.


GaryAustin1:
Using the nextjs example code and setting my token timeout on the server to 100 seconds a loop updating the screen from the database fails right after 100 seconds with 401 {"message":"JWT expired"} errors. 

I was discussing this issue with @silentworks in his work on Sveltekit version and decided to test here.

It is not clear how auth-helpers refreshes the token is the user is on the same page for more than the expire time.  Examples of this could be stock ticker type applications, a main screen with table rows listed that has user search filter or a refresh icon.  The user can interact with this page, leave for an hour and come back and the token will expire.  Auth helpers turns off the refresh time in the Supabase client and does not replace it with some way to keep token refreshed if on same page.

Sitting in main example screen with a 20 second loop reading a table row and updating screen.
![image](https://user-images.githubusercontent.com/54564956/163730154-405db65a-2926-4c92-af25-00ab362299a5.png)

I modified index.ts in example to do this loop.  Only change.
```
useEffect(() => {
    async function loadData() {
        let intervalID;
		function repeat() {
			intervalID = setInterval(repeatUpdate, 20000);
		}

		async function repeatUpdate() {
			console.log('20 seconds elapsed.');
			const { data } = await supabaseClient.from('test').select('*').eq('id',63).single();
				console.log(data)
			setData( data);
		}
		repeat()


     // const { data } = await supabaseClient.from('test').select('*').eq('id',63).single();
     // setData(data);
    }
    if (user) loadData();
```

GaryAustin1:
Adding a 2nd worse case...
Using the base sample code with just a single database load on index page...
I toggle between Home and withAuthRequired link.  After 100 seconds the select to the server fails with jwt expired also.  So no refreshing of token is being done between route changes.

![image](https://user-images.githubusercontent.com/54564956/163734627-adaa61f2-36b6-470a-95ea-fae6988eaaef.png)

Note: I was getting double console.logs which I have seen before sometimes in dev environment with React back when I used.

dabarrell:
I'm facing this issue as well. I've been reading through the repo to see what would need to change, and it seems like 
a) setting `autoRefreshToken` to `true` in the client; and
b) handling `TOKEN_REFRESHED` here in the callback handler (https://github.com/supabase-community/supabase-auth-helpers/blob/50709164dd2ef83990dd1b0456649f417aed3fd9/src/nextjs/handlers/callback.ts#L29)

[...]

Issue #35, opened 2022-03-10
Refresh token when page becomes visible


thorwebdev:
https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event

This is something we can consider. It might help in scenarios where the page has been in the background for a long while and the token has expired. Using the visibility change event we can detect the page coming visible again and can refresh the token if needed before the user starts interacting with the page.

thorwebdev:
https://github.com/supabase-community/supabase-auth-helpers/pull/92

PR #78, opened 2022-04-19
Automatically refresh access token before it expires

dabarrell:
Addresses https://github.com/supabase-community/supabase-auth-helpers/issues/73

Uses a similar timeout mechanism to `gotrue-js`'s `autoRefreshToken` function, but simplified.

With the `refreshToken` available in the frontend after the first commit, it could theoretically be passed to the `supabaseClient.auth` instance so that we can use its `autoRefreshToken` function, but the only current interface for setting the `refreshToken` is `supabaseClient.auth.setSession(refreshToken)`, which triggers a refresh of the token, which we don't want. So the (arguably cleaner) alternative is to handle the refresh within `UserProvider`.

### DIFF
--- a/src/react/components/UserProvider.tsx
+++ b/src/react/components/UserProvider.tsx
@@ -1,12 +1,13 @@
-import React, {
-  useEffect,
-  useState,
-  createContext,
-  useContext,
-  useCallback
-} from 'react';
-import { useRouter } from 'next/router';
 import { SupabaseClient, User } from '@supabase/supabase-js';
+import { useRouter } from 'next/router';
+import React, {
+  createContext,
+  useCallback,
+  useContext,
+  useEffect,
+  useRef,
+  useState
+} from 'react';
 
 export type UserState = {
   user: User | null;
@@ -17,12 +18,20 @@ export type UserState = {
 
 const UserContext = createContext<UserState | undefined>(undefined);
 
-type UserFetcher = (
-  url: string
-) => Promise<{ user: User | null; accessToken: string | null }>;
+type UserFetcher = (url: string) => Promise<{
+  user: User | null;
+  accessToken: string | null;
+  refreshToken: string | null;
+  expiresAt: number | null;
+}>;
 const userFetcher: UserFetcher = async (url) => {
   const response = await fetch(url);
   return response.ok ? response.json() : { user: null, accessToken: null };
+};
+
+type RefreshToken = {
+  token: string;
+  expiresAt: number;
 };
 
 export interface Props {
@@ -46,14 +55,21 @@ export const UserProvider = (props: Props) => {
   const [accessToken, setAccessToken] = useState<string | null>(null);
   const [isLoading, setIsLoading] = useState<boolean>(!initialUser);
   const [error, setError] = useState<Error>();
+  const [refreshToken, setRefreshToken] = useState<RefreshToken | null>(null);
+  const refreshTimer = useRef<NodeJS.Timeout | undefined>();
   const { pathname } = useRouter();
 
   const checkSession = useCallback(async (): Promise<void> => {
     try {
-      const { user, accessToken } = await fetcher(profileUrl);
+      const { user, accessToken, refreshToken, expiresAt } = await fetcher(
+        profileUrl
+      );
       if (accessToken) {
         supabaseClient.auth.setAuth(accessToken);
         setAccessToken(accessToken);
+      }
+      if (refreshToken && expiresAt) {
+        setRefreshToken({ token: refreshToken, expiresAt });
       }
       setUser(user);
       if (!user) setIsLoading(false);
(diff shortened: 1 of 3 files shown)


src/react/components/UserProvider.tsx
```diff
@@ -46,22 +55,56 @@ export const UserProvider = (props: Props) => {
...
+      );
       if (accessToken) {
         supabaseClient.auth.setAuth(accessToken);
         setAccessToken(accessToken);
       }
+      if (refreshToken && expiresAt) {
+        setRefreshToken({ token: refreshToken, expiresAt });
+      }
       setUser(user);
       if (!user) setIsLoading(false);
```
dabarrell, line 75: Seems like there's some double-handling of `isLoading` in this component – it's set in `checkSession`, and in the two `useEffect` hooks that call `checkSession`. Seems worth cleaning up, but I didn't want to muddy this PR.

dabarrell: While I'm thinking about it, it seems strange to set `isLoading` to `true` on subsequent reloads of the user (after the first load). If you think about how `isLoading` would be used practically, it'd be to show loading states etc on an initial page load. As a consumer of the hook, I wouldn't expect `isLoading` to then flick back to false then true again every time the user is refreshed – it'd cause loading states to briefly pop back in, which would be strange

thorwebdev:
Thanks for this. I think the setTimeout itself, isn't enough though. Maybe needs to be paired with something like https://github.com/supabase-community/supabase-auth-helpers/issues/35 ?

Also, we shouldn't refresh the token on the client-side, but rather on the server-since the cookies are the source of truth and we don't want them to become out of sync. So rather than calling `supabaseClient.auth.setSession` we should make a get request to `api/user` which then uses `getUser` to refresh the token. `getUser` should be the only method that handles token refreshes. Wdyt, does that make sense?

dabarrell:
`setTimeout` feels a little thin to me, but I took inspiration from `@supabase/gotrue-js` which uses `setTimeout` for its auto refresh mechanism. I think the issue with https://github.com/supabase-community/supabase-auth-helpers/issues/35 is that it allows the token to expire, meaning if the user comes back to the page after it's expired, there's a chance that requests might be sent with an expired token before the refresh is completed.

As an example, my app uses SWR, which I've configured to refresh data when the page becomes visible again (as is a common pattern). This would mean the page data refresh and the token refresh would happen at the same time, and the former would fail.

>Also, we shouldn't refresh the token on the client-side, but rather on the server-since the cookies are the source of truth and we don't want them to become out of sync.

With this PR, they won't get out of sync as calling `supabaseClient.auth.setSession` triggers an event, which is picked up by `supabaseClient.auth.onAuthStateChange` in `UserProvider` and pushed to the `/api/auth/callback` route, which updates the cookies with the new tokens.

That said, I could understand the design choice to avoid reading refresh tokens in the client, and instead let the handlers pull them out of cookies and complete the refresh there. As it stands, calling `/api/auth/user` won't cause a refresh unless the access token has already expired, but perhaps we could add a `/api/auth/refresh` handler specifically for this use-case?

dabarrell:
> the cookies are the source of truth

I'm not sure this PR breaks that principle (which I agree with) – it's reading the refresh token from the cookies then using it in the frontend in the same way that the access token is being used. It's not being updated elsewhere - only in the cookies

dabarrell:
>Maybe needs to be paired with something like https://github.com/supabase-community/supabase-auth-helpers/issues/35 ?

Apologies! I misread this initially – I see now that you're suggesting using both `setTimeout` and also something like #35. That could certainly make it more reliable!

silentworks:
I think @GaryAustin1 has a more robust fix as its at the gotrue-js level https://github.com/supabase/supabase/issues/6464

thorwebdev:
@silentworks, AFAICS @GaryAustin1 solution is for client-side refreshes when the tokens are managed in localStorage. With the auth helpers, token management is moved to the server-side via cookies, so the refresh would need to happen there so that tokens don't get out of sync. 

Since we're moving away from storing tokens in localStorage in the future, I'd prefer to have the auth helpers not rely on this part of `gotrue-js` if that makes sense?

GaryAustin1:
All, 
This is a multifaceted issue. 
1) detection of expiring toke should be a common util or at least formula for all of Supabase.  Currently there is no specified minimum token time.  It would help if it was something like two minutes.  Then a 10 sec number could be used reasonably.  A formula could also be used based on the exp life provided back from gotrue.  In the end this refresh is just to get thru fixed immediate code though.

2) A reliable means of refreshing the token that does not depend on the timers is needed. I highly recommend Supabase.js check the token expiration when it sets the bearer token on each call. If about to expire it can refresh it and generate an event that can be monitored to sync severside/local copies. This is the method I am attempting to simulate with fix 1 in my code.  @thorwebdev I’m not sure my report in Supabase has anything to do with local storage.  It is a timer issue on mobile that will affect any client side timer and storage method of the session data. Even auth helpers apps will potentially not make a serverside call for hours. 

3). All testing should be done with low token expire time set on server and robust use of RLS database calls on both server side and client side.

thorwebdev:
Yes, it's happening on a timer because of realtime. See some suggested improvements here: https://github.com/supabase/gotrue-js/pull/274

thorwebdev:
This has been added in https://github.com/supabase-community/supabase-auth-helpers/pull/92

Issue #81, opened 2022-04-22
Cannot install @supabase/auth-helpers-nextjs error not in the npm registry


Dipeshwagle:
When I try to install `@supabase/auth-helpers-nextjs`  I get the following error message.

```
npm install @supabase/auth-helpers-nextjs
npm ERR! code E404
npm ERR! 404 Not Found - GET https://registry.npmjs.org/@supabase%2fauth-helpers-nextjs - Not found
npm ERR! 404
npm ERR! 404  '@supabase/auth-helpers-nextjs@latest' is not in the npm registry.
npm ERR! 404 You should bug the author to publish it (or use the name yourself!)
npm ERR! 404
npm ERR! 404 Note that you can also install from a
npm ERR! 404 tarball, folder, http url, or git url.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/dipeshwagle/.npm/_logs/2022-04-22T12_19_44_665Z-debug.log

```

bill-offertrackr:
You'll want to install `@supabase/supabase-auth-helpers`.  The nextjs utils can then be imported like this:

```
import { withAuthRequired } from '@supabase/supabase-auth-helpers/nextjs';
```

okezieuc:
I created a PR to update this in the README at #87.

thorwebdev:
I've changed back to default to the `main` branch to make sure you see what's currently released.

PR #87, opened 2022-04-30
update README on installing Supabase Next.js auth helpers

okezieuc:
This updates the README on installing the Next.js auth helpers based on the discussion at #81.


silentworks:
We are moving to new package names and what we have is correct as it stands, we should probably just add in the README that the old package repo is on the `main` branch and not the `next` branch.

PR #88, opened 2022-05-02
using new lib

emersonlaurentino:
On Next.js README use a lib that does not exist, so I update to new lib (that works).


emersonlaurentino:
Why do you close @thorwebdev?

thorwebdev:
Because the next branch will be under the new namespace that's in there. So the PR isn't valid. Hope that makes sense.

emersonlaurentino:
Yep, makes sense, I question because you just closed the PR without any context, thanks for explaining.

Issue #103, opened 2022-05-21
Support to require certain roles using nextjs middleware


mr-bjerre:
Currently we can do the do the following to protect routes via middleware

```typescript
import { withMiddlewareAuth } from "@supabase/supabase-auth-helpers/nextjs/middleware";

export const middleware = withMiddlewareAuth({ redirectTo: "/login" });
```

However I'd like to restrict certain routes for certain roles; e.g. admins only. I keep track of my admins in a table so it is basically a question of reading data from the database and require some condition based on that. Any suggestion on how to achieve that?

mr-bjerre:
We could add an option to provide a user callback to determine if a user is allowed or not.

So extend the parameter options to this

```typescript
export interface withMiddlewareAuthOptions {
  redirectTo?: string;
  cookieOptions?: CookieOptions;
  tokenRefreshMargin?: number;
  permissionGuard?: {
    isPermitted: (user: User) => boolean
    redirectTo: string;
  }
}
```

And then add the following line before returning

```typescript
if (permissionGuard && permissionGuard.isPermitted(authResult.user)) {
    throw new Error('User is not permitted, redirecting');
}
```

We then have to update the error handling to handle two different type of errors (since it needs to redirect to different paths).

If you think this is viable then I'll put together a PR - let me know.

ferpintado:
This is a case that I am running into, will be very helpful for sure

mr-bjerre:
Simarily we could do something similar on [`withApiAuth`](https://github.com/supabase-community/supabase-auth-helpers/blob/main/src/nextjs/README.md#protecting-api-routes). I'm waiting for the maintainers to comment/accept #104  before "copying" the solution over.

silentworks:
PR for this was merged and will be out in the next release.

PR #104, opened 2022-05-22
Add permission guard to midddleware

mr-bjerre:
Add new feature requested in #103

### DIFF
--- a/src/nextjs/middleware/withMiddlewareAuth.ts
+++ b/src/nextjs/middleware/withMiddlewareAuth.ts
@@ -11,6 +11,14 @@ import {
   NextResponseAdapter
 } from '../../shared/adapters/NextMiddlewareAdapter';
 import { jwtDecoder } from '../../shared/utils/jwt';
+import type { User } from '@supabase/supabase-js';
+
+class NoPermissionError extends Error {
+  constructor(message: string) {
+    super(message);
+    Object.setPrototypeOf(this, NoPermissionError.prototype);
+  }
+}
 
 export interface withMiddlewareAuthOptions {
   /**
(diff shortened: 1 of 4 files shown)


src/nextjs/middleware/withMiddlewareAuth.ts
```diff
@@ -119,12 +131,20 @@ export const withMiddlewareAuth: withMiddlewareAuth =
...
+        throw new NoPermissionError('User is not permitted, redirecting');
       }
 
       // Authentication successful, forward request to protected route
       return res;
     } catch (err: unknown) {
-      const { redirectTo = '/' } = options;
+      let { redirectTo = '/' } = options;
+      if (err instanceof NoPermissionError && !!options.permissionGuard) {
+        redirectTo = options.permissionGuard.redirectTo;
```
ferpintado, line 146: I'd default to `redirectTo` from options if this is not defined.

mr-bjerre: Yeah good idea. I've added that now

ferpintado (approved):

ferpintado (approved):

silentworks:
This PR looks specific to your use case and probably won't work for general use. I can see you are referencing an `admin` table which would be user created.

mr-bjerre:
> This PR looks specific to your use case and probably won't work for general use. I can see you are referencing an `admin` table which would be user created. 

No this is not specific to my use case. I just added the admin table as an example. It could just as well be `user => user.email.endswith("mycompany.com")`.

I'd argue that this solves a _very_ common use case. Ie. making it possible to restrict access to certain roles, users etc. Instead of a binary response - logged in or not. With this you have role based access via middleware.

ferpintado:
I do find it useful as well. In my case, I need to check `user.user_metadata.role` to know whether the user has permissions. 

I do agree, having the example in `examples/nextjs/pages/middleware-protected/_middleware.ts` is very specific. Perhaps just needs to be added in the README.

mr-bjerre:
I'll update example and docs later tonight

silentworks:
@mr-bjerre I know this is a big ask, but can you make this change to the new `main` branch, please?

mr-bjerre:
> @mr-bjerre I know this is a big ask, but can you make this change to the new `main` branch, please?

Opened new PR #126 in favor of this

PR #108, opened 2022-05-30
[WIP] Add SvelteKit support to the auth helpers

silentworks:


examples/sveltekit/example.env
```diff
@@ -0,0 +1,3 @@
+# Update these with your Supabase details from your project settings > API
```
thorwebdev, line 1: nit: I'd call this `.env.example` and maybe add instructions to readme to copy it into `.env` file?

thorwebdev (approved):
LGTM 🥳

PR #115, opened 2022-06-01
Update shared to be its own npm package

silentworks:
Add shared as a dependency of next, react, svelte and sveltekit packages


examples/sveltekit/src/routes/__layout.svelte
```diff
@@ -2,13 +2,13 @@
 	import { goto } from '$app/navigation';
 	import { session } from '$app/stores';
 	import { supabaseClient } from '$lib/db';
-	import { AuthHelper } from '@supabase/auth-helpers-svelte';
+	import { SuperAuthHelper } from '@supabase/auth-helpers-svelte';
```
thorwebdev, line 5: Should this be `SupaAuthHelper`?

packages/shared/package.json
```diff
@@ -1,21 +1,24 @@
...
     "next": "^12.1.5",
     "react": ">=17.0.2 <18.0.0 || >=18.0.0-0 <19.0.0",
-    "react-dom": "^17.0.2 || ^18.0.0-0"
-  },
-  "peerDependencies": {
-    "react": ">=17.0.2 <18.0.0 || >=18.0.0-0 <19.0.0",
-    "react-dom": "^17.0.2 || ^18.0.0-0"
+    "react-dom": "^17.0.2 || ^18.0.0-0",
+    "tsconfig": "workspace:*",
+    "tsup": "^5.12.5"
```
thorwebdev, line 22: With this setup we should be able to move the `jwtDecoder` back into shared and have `jose` be a dependency here, right?

packages/sveltekit/package.json
```diff
@@ -38,15 +36,15 @@
...
     "svelte": "^3.48.0",
     "svelte2tsx": "^0.5.10",
     "tslib": "^2.3.1",
     "typescript": "^4.7.2"
   },
   "dependencies": {
+    "@supabase/auth-helpers-shared": "workspace:*",
     "@supabase/supabase-js": "^1.35.3",
-    "del-cli": "^4.0.1",
     "jose": "^4.8.1"
```
thorwebdev, line 48: should be able to move `jose` to shared now, right?

thorwebdev (changes requested):
Let's move the JWT stuff (including the `jose` dependency) back into shared.

examples/sveltekit/src/routes/__layout.svelte
silentworks: I couldn't think of a good name, open to suggestions.

thorwebdev: I think `AuthHelper` works, or `SupaAuthHelper` to stick with the brand 👍

thorwebdev (approved):
LGTM, thanks!

PR #146, opened 2022-06-26
SvelteKit: Fix supabaseServerClient with request and withApiAuth return type

johnathanludwig:
When a request is passed in to `supabaseServerClient` it was looking at the header for the access token. This pulls the access token off of the cookie header.

When using `withApiAuth`, the returned type was hardcoded to `{}`. This allows the proper types to be inferred from the function so they may be used downstream. See this [TS Playground](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAWxgZzTMBzAKgTwAcBTACmDAC5FSBKAXgD5EBvAX1pYChFfEAnYlBD8kFOgG4ubLl1CRYCRFmDF+ggCYESAHlyNyVGg2a5OPPoOGjE42lJlzw0eEijE0UOqwu8rIpBZgODhqAHIARgAmAGYwgBpEACMAQ35wgBYAVgA2MJlHCARPAQ9CYuJEehR0TBxtMndPWlkisBLBNHL24iiq5TBVdWItIkaPKFogA) for an example.

### DIFF
--- a/packages/sveltekit/src/utils/supabaseServerClient.ts
+++ b/packages/sveltekit/src/utils/supabaseServerClient.ts
@@ -1,6 +1,10 @@
 import type { SupabaseClient } from '@supabase/supabase-js';
 import { skHelper } from '../instance';
-import { type CookieOptions, COOKIE_OPTIONS } from '@supabase/auth-helpers-shared';
+import {
+  type CookieOptions,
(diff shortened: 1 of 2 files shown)


silentworks (approved):

silentworks:
Thank you for this PR.

PR #163, opened 2022-07-07
sveltekit: Create supabase endpoints in handle hooks

david-plugge:
This removes the need for user defined endpoints.


jycouet:
Nice work @david-plugge 

Once this is completed, I would like to use it https://github.com/jycouet/supabase-graphql-example (and upgrade with our latest stuff from GraphQL 😎)

You wanna participate? Ping me 👍

david-plugge:
To prevent users from creating routes that are handled by this package a warning will be logged if the response status is not 404 or 405.
```js
if (!(res.status === 405 || res.status === 404)) {
  logger.warn(
    `@supabase/auth-helpers-sveltekit handles the route '${endpointPath}'`
  );
}
```

I´m not very convinced by this as the user may return a 404 in his endpoint.
We could also completely ignore sveltekits handlers and just dont call `resolve(event)`.

packages/sveltekit/src/handlers/user.ts
```diff
@@ -1,108 +1,45 @@
...
+      logger.warn(
+        `@supabase/auth-helpers-sveltekit handles the route '${endpointPath}'`
+      );
+    }
 
-      event.locals.user = null;
-      event.locals.accessToken = null;
-      return await resolve(event);
+    // check request method
+    if (req.method !== 'GET') {
```
silentworks, line 29: This should be POST and not GET.

packages/sveltekit/src/handlers/user.ts
```diff
@@ -1,108 +1,45 @@
...
+      const headers = new Headers({
+        Allow: 'GET'
```
silentworks, line 31: Allowing only POST requests not GET requests.

packages/svelte/src/SupaAuthHelper.svelte
```diff
@@ -9,8 +9,9 @@
 
   // Props
   export let supabaseClient: SupabaseClient;
-  export let callbackUrl = '/api/auth/callback';
-  export let profileUrl = '/api/auth/user';
+  export let endpointPrefix = '/api/auth';
```
silentworks, line 12: It's probably worth pulling thru the `ENDPOINT_PREFIX` const from the `@supabase/auth-helpers-shared` package too.

silentworks (changes requested):

silentworks:
The callback handler is going to return a 404 because the route doesn't exists and the handler is going to return the status code from that request. We would need to return a new Response with a 200 status code, but this would need to contain the cookie header from the previous response.

silentworks (approved):

david-plugge:
@silentworks the packages are not updated, is that intended?

PR #170, opened 2022-07-10
chore: next 12.2.x middleware upgrade

boredland:
closes https://github.com/supabase-community/auth-helpers/issues/168
closes https://github.com/supabase-community/auth-helpers/issues/144
likely closes https://github.com/supabase-community/auth-helpers/issues/166

my general idea for the middleware part would be, that we'll be able to define a list of route guards, each may have a list of path-matchers (similar to the stock matcher configuration) and each may have the options we were defining before with the nested routes.

regarding the matchers I have yet to find a nice approach on how to replicate the behavior of the [stock matchers](https://nextjs.org/docs/advanced-features/middleware#matcher). So, right now this is meant as a draft to get some feedback if that API would be sth. we could live & work with.

### DIFF
--- a/examples/nextjs/middleware.ts
+++ b/examples/nextjs/middleware.ts
@@ -1,9 +1,10 @@
 import { withMiddlewareAuth } from '@supabase/auth-helpers-nextjs/dist/middleware';
 
-export const middleware = withMiddlewareAuth({
-  redirectTo: '/login',
+export const middleware = withMiddlewareAuth([{
+  redirectTo: '/',
+  matcher: ['/middleware-protected'],
   authGuard: {
     isPermitted: async (user) => user.email?.endsWith('@example.com') ?? false,
     redirectTo: '/insufficient-permissions'
   }
-});
+}]);
--- a/examples/nextjs/middleware.ts
+++ b/examples/nextjs/middleware.ts
@@ -1,4 +1,4 @@
-import { withMiddlewareAuth } from '@supabase/auth-helpers-nextjs/dist/middleware';
+import { withMiddlewareAuth } from '@supabase/auth-helpers-nextjs/middleware';
 
 export const middleware = withMiddlewareAuth([{
   redirectTo: '/',
--- a/examples/nextjs/middleware.ts
+++ b/examples/nextjs/middleware.ts
@@ -2,9 +2,12 @@ import { withMiddlewareAuth } from '@supabase/auth-helpers-nextjs/dist/middlewar
 
 export const middleware = withMiddlewareAuth([{
   redirectTo: '/',
-  matcher: ['/middleware-protected'],
+  matcher: ['/middleware-protected/:path*'],
   authGuard: {
     isPermitted: async (user) => user.email?.endsWith('@example.com') ?? false,
     redirectTo: '/insufficient-permissions'
   }
+}, {
+  redirectTo: '/',
+  matcher: ['/middleware-protected-page'],
 }]);
--- a/packages/nextjs/src/utils/supabaseServerClient.ts
+++ b/packages/nextjs/src/utils/supabaseServerClient.ts
@@ -38,6 +38,7 @@ export default function supabaseServerClient(
   }
   const access_token =
     context.req.cookies[`${cookieOptions.name}-access-token`];
-  supabaseClient.auth.setAuth(access_token);
+  if (access_token)
+    supabaseClient.auth.setAuth(access_token);
   return supabaseClient;
 }
--- a/packages/nextjs/src/utils/supabaseServerClient.ts
+++ b/packages/nextjs/src/utils/supabaseServerClient.ts
@@ -1,7 +1,6 @@
 import { GetServerSidePropsContext, NextApiRequest } from 'next';
 import { supabaseClient, SupabaseClient } from './initSupabase';
 import { CookieOptions } from '@supabase/auth-helpers-shared';
-import { NextRequest } from 'next/server';
 
 /**
  * This is a helper method to wrap your SupabaseClient to inject a user's access_token to make use of RLS on the server side.
--- a/packages/nextjs/src/utils/supabaseServerClient.ts
+++ b/packages/nextjs/src/utils/supabaseServerClient.ts
@@ -1,6 +1,7 @@
 import { GetServerSidePropsContext, NextApiRequest } from 'next';
 import { supabaseClient, SupabaseClient } from './initSupabase';
 import { CookieOptions } from '@supabase/auth-helpers-shared';
+import { NextRequest } from 'next/server';
 
 /**
  * This is a helper method to wrap your SupabaseClient to inject a user's access_token to make use of RLS on the server side.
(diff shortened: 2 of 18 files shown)


boredland:
edit: I guess nextjs processes paths extremely similar to https://www.npmjs.com/package/url-pattern, so I'll just add that for now. Better ideas (preferrably using next utils) welcome!

packages/nextjs/src/utils/supabaseServerClient.ts
```diff
@@ -38,6 +38,7 @@ export default function supabaseServerClient(
   }
   const access_token =
     context.req.cookies[`${cookieOptions.name}-access-token`];
```
Cryrivers, line 40:
```suggestion
  const { value: access_token } =
    context.req.cookies.getWithOptions(`${cookieOptions.name}-access-token`);
```

packages/shared/src/adapters/NextMiddlewareAdapter.ts
```diff
@@ -8,7 +8,7 @@ export class NextRequestAdapter implements RequestAdapter {
   }
 
   setRequestCookie(name: string, value: string) {
-    this.req.cookies[name] = value;
+    this.req.cookies.set(name, value);
   }
```
Cryrivers, line 12: I guess this line should also apply to `[NextAdapter.ts](https://github.com/boredland/auth-helpers/blob/chore/next-middleware-upgrade/packages/shared/src/adapters/NextAdapter.ts)`

packages/nextjs/src/utils/supabaseServerClient.ts
Cryrivers: According to https://nextjs.org/docs/messages/middleware-upgrade-guide#cookies-api-revamped

packages/shared/src/adapters/NextMiddlewareAdapter.ts
boredland: sadly no, the cookies property in NextApiRequest is just an object:
```
cookies: Partial<{
        [key: string]: string;
    }>;
```

jonathanwilke (approved):

jonathanwilke:
This would be really nice if it could be merged, since I can't get my app to work with the latest version of NextJS.

silentworks:
@boredland is this ready to go? I've had a look over the code and it looks good to me.

boredland:
> @boredland is this ready to go? I've had a look over the code and it looks good to me.

I'd say so yes. Added an example app so you can try it out...

413n (approved):
Really nicely done!

413n:
Hey @silentworks, did you found the time to test the example app on this PR?
If yes or if that's not necessary, I think @boredland can then proceed to merge it!

boredland:
> I think @boredland can then proceed to merge it!

I think he can't, because "At least 1 approving review is required by reviewers with write access".

413n:
Yeah I meant that if @silentworks approves it then you can merge it!

packages/nextjs/src/utils/supabaseServerClient.ts
```diff
@@ -36,8 +36,9 @@ export default function supabaseServerClient(
   if (!context.req.cookies) {
     return supabaseClient;
   }
-  const access_token =
-    context.req.cookies[`${cookieOptions.name}-access-token`];
-  supabaseClient.auth.setAuth(access_token);
+  const { value: access_token } =
+    context.req.cookies.getWithOptions(`${cookieOptions.name}-access-token`);
```
silentworks, line 40: This fails on the build, is this function name correct?

silentworks (changes requested):

silentworks:
Please run `pnpm build:nextjs` after making these changes from the root directory. Also run the example app to see that signing in and signing out is working. Currently the url isn't redirecting properly when I tested this and the hash with the access_token is in the URL when I tried signing in.

examples/nextjs/middleware.ts
```diff
@@ -1,9 +1,13 @@
-import { withMiddlewareAuth } from '@supabase/auth-helpers-nextjs/dist/middleware';
+import { withMiddlewareAuth } from '@supabase/auth-helpers-nextjs/middleware';
```
silentworks, line 1: I'm not sure if this is going to work, because this will lead to a warning about `jose`. 
```
warn  - ../../node_modules/.pnpm/jose@4.8.1/node_modules/jose/dist/browser/runtime/env.js
A Node.js API is used (process.versions at line: 6) which is not supported in the Edge Runtime.
Learn more: https://nextjs.org/docs/api-reference/edge-runtime
```
This is the reason why we have `/dist/` in the import path.

boredland: with the dist import I think there may be something odd about the import path too, as I'm hitting `Module '"@supabase/auth-helpers-nextjs/dist/middleware"' has no exported member 'withMiddlewareAuth'.ts(2305)`

boredland: ah, I just saw that I added an [import configuration](https://github.com/supabase-community/auth-helpers/pull/170/files#diff-e5237f683dda3354b835c7c7c94b9759db2c743d4ba94d47d7f8b8e0b2bfb442R30-R49) for that to the package.json (this PR got really stale for me over the weeks). How are you getting that error?

boredland:
> Currently the url isn't redirecting properly when I tested this and the hash with the access_token is in the URL when I tried signing in.

While you're right, I am not so sure this is connected to this PR, as this shouldn't be connected to middleware, should it?

boredland:
> Currently the url isn't redirecting properly when I tested this and the hash with the access_token is in the URL when I tried signing in.

While you're right, I am not so sure this is connected to this PR, as this shouldn't be connected to middleware, should it?

silentworks:
> > Currently the url isn't redirecting properly when I tested this and the hash with the access_token is in the URL when I tried signing in.
> 
> While you're right, I am not so sure this is connected to this PR, as this shouldn't be connected to middleware, should it?

I think this is related to the PR somehow, not sure which if it's a part of the NextJS 12.2 change or something else. When I test the main branch this issue doesn't exist at all.

examples/nextjs/middleware.ts
silentworks: By just running `pnpm dev --filter=@example/nextjs` then opening the site in the browser and you will see the error in the terminal.

silentworks: > with the dist import I think there may be something odd about the import path too, as I'm hitting `Module '"@supabase/auth-helpers-nextjs/dist/middleware"' has no exported member 'withMiddlewareAuth'.ts(2305)`

You will only get this if you leave the [import configuration](https://github.com/supabase-community/auth-helpers/pull/170/files#diff-e5237f683dda3354b835c7c7c94b9759db2c743d4ba94d47d7f8b8e0b2bfb442R30-R49) in, but if you remove that it is no longer there.

thorwebdev:
@boredland thanks so much for this, and sorry for moving so slowly here 🙈 

Two things. We're working on a new approach on the `next` branch which incorporates the changes of supabase-js V2 RC. In the meantime we'll release https://github.com/supabase/auth-helpers/pull/227 as v `0.2.8` to bridge the gap until we can release the `next` changes. 

Sorry again that this didn't work out, but we really appreciate your contribution and helping us nudge in the right direction! 💚

PR #189, opened 2022-07-20
Implicit typing to endpoint function parameters

dlugoschvincent:
Adding type to endpoint functions in sveltekit email password example to enable implicit typing of parameters. Also more consistent with documentation.


silentworks:
This was included as part of #190.

PR #214, opened 2022-08-06
SvelteKit `withApiAuth/withPageAuth` nullable `user` arg

probablykasper:
It seems like `withApiAuth` and `withPageAuth` are meant to handle cases where the `user` argument is null, so this PR makes them nullable


david-plugge:
Should be solved now

PR #239, opened 2022-08-23
chore: upgrade to supabase-js v2

alaister:

Upgrades to the `rc` release of `gotrue-js`

## Additional context

At this stage, this can be treated as a proof-of-concept of a technically different approach to the `auth-helpers`.

Refreshing behaviour is now handled by `gotrue-js`, and the session is synced by using shared cookies.

The `nextjs` example has been updated to work with this new approach. If we decide this is the direction we want to take, we can look at upgrading the other libraries/examples too.


dukesx:
Looks promising but we really need a better way to manage auth state. OAuth is one weak point of Supabase auth, especially when it comes to redirecting to index page and you immediately need the session state.

grantvine-fg:
Will this fix the "getting logged out with a stale token" issue I have ?

packages/shared/src/types.ts
```diff
@@ -1,30 +1,6 @@
...
-  // (Optional) The Cookie name prefix. Defaults to `sb` meaning the cookies will be `sb-access-token` and `sb-refresh-token`.
-  name?: string;
-  // (Optional) The cookie lifetime (expiration) in seconds. Set to 8 hours by default.
-  lifetime?: number;
-  // (Optional) The cookie domain this should run on. Leave it blank to restrict it to your domain.
-  domain?: string;
-  path?: string;
-  // (Optional) SameSite configuration for the session cookie. Defaults to 'lax', but can be changed to 'strict' or 'none'. Set it to false if you want to disable the SameSite setting.
-  sameSite?: string;
-}
```
AaronDewes, line 28: I just tried this version, and this change broke builds or me, because the sveltekit helpers seemed to import it.

silentworks: SvelteKit helpers haven't been upgraded to work with v2 as yet.

packages/shared/src/supabase-browser.ts
```diff
@@ -0,0 +1,72 @@
+import { createClient, Session } from '@supabase/supabase-js';
+import { parse, serialize } from 'cookie';
+import { CookieOptions } from './types';
+import { isBrowser } from './utils/helpers';
+
+export function createBrowserSupabaseClient({
```
nickbrinser, line 6: With the new v2 createClient API shouldn't we be expecting a users to provide a Database type?

PR #245, opened 2022-09-11
sveltekit helpers update

### LINKED ISSUE #230: SvelteKit removed Session store
# Bug report

## Describe the bug

The latest version of SvelteKit removes the Session store, breaking the Svelte and SvelteKit auth helpers.
See: https://github.com/sveltejs/kit/discussions/5883#discussion-4302953

## Expected behavior

Be able to use auth-helpers with latest version of SvelteKit

david-plugge:
Closes #230

Support new versions of sveltekit (around 1.0.0-next.440 upwards, random guess based on their changelog)

We can´t use auth helpers with the current version of sveltekit.

## Additional context

Since there is no need for a Svelte component we could build the sveltekit helpers with tsup like the other packages.

PS:
SvelteKit is in release candidate phase!

Edit: very sorry for this huge commit :/


dlugoschvincent:
I have one question. I don't see any example or possibility to trigger an action with authentication. All I see is authenticated data retrieval, but not authenticated data posting. Am I seeing this correctly, or how would I go about making an authenticated insert/update/delete from the server side?

david-plugge:
Using [`supabaseServerClient`](https://github.com/supabase/auth-helpers/pull/245/commits/31982a6be46c67fa22a7274a8296841b5226ed2c#diff-3de7d9878d6153001634521c738eb846353b5c6155a1a636b8fc02b5959c3b83) you get a supabase instance that sends the users accessToken on every request.

I´ll add an example though!

david-plugge:
To clarify a bit:

The helpers (load, serverLoad, server) check if there is a logged in user and if not redirects with the status and location provided. So the callback only runs when there is a session, which makes sense to me. There won´t be many pages that check if there is no session right?

There could be added even more stuff than just a session to the event.
```ts
export const load: PageLoad = loadWithSession(
	{ status: 303, location: '/' },     	// the user is redirected if there is no session
	async ({ session }) => {				// we have a session and can make authenticated requests
		const { data: testTable } = await supabaseServerClient(session.accessToken)
			.from<TestTable>('test')
			.select('*');
		return {
			testTable,
			user: session.user  // the user is correctly typed, no need to check if it´s nullish
		};
	}
);
```
Since the helper has access to the session this is possible aswell:
```ts
export const load: PageLoad = loadWithSession(
	{ status: 303, location: '/' }, 
	async ({ getSupabaseClient }) => {
		const { data: testTable } = await getSupabaseClient()
			.from<TestTable>('test')
			.select('*');
		return {
			testTable,
        };
	}
);
```

Thoughts?

dlugoschvincent:
Ok, thank you, I figured as much, but how do I get the access token in an action as there is a loadWithSession function but no “action with session” function. Maybe I am confused.

Imagine a scenario like this, where I want to make sure only authenticated users can insert into the database.
Would something like this be possible?

```typescript
export const actions: Actions = {
  add: async ({ request }) => {
    const json = await request.json();
    supabaseServerClient(json.accessToken)
      .from<definitions['Food']>('Food')
      .insert({ bar_code: json.barCode });
  }
}
```

david-plugge:
You can use `locals` in actions
```ts
export const actions: Actions = {
  add: async ({ request, locals }) => {
    if (!locals.accessToken) {
        throw redirect(303, '/signin');
    }
    const json = await request.json();
    supabaseServerClient(locals.accessToken)
      .from<definitions['Food']>('Food')
      .insert({ bar_code: json.barCode });
  }
}
```

dlugoschvincent:
Okay, thank you so much, I will give it a try! Sorry for taking your time. Having an example for actions would be really helpful. I could think one up once this PR gets merged.

esojrafael:
Thank you good sir

david-plugge:
[Updated docs](https://github.com/david-plugge/supabase-auth-helpers/blob/sveltekit-new-api/packages/sveltekit/README.md)

Feel free to leave some comments

kvetoslavnovak:
I guess this [dangerous SvelteKit data leaking in SSR issue](https://github.com/sveltejs/kit/discussions/4339) should be checked as well.
There is even a [proposed solution](https://github.com/sveltejs/kit/discussions/4339#discussioncomment-3258927) for a custom store reinventing the depreciated session store.

david-plugge:
Good catch, i found a few more issues in the readme.
Since we are only using the `page` store there is no chance data can leak.

david-plugge:
I noticed one problem though that is also included in the current version.
The `supabaseServerClient()` method returns the supabase instance and sets the accessToken. If there is no token it wont be set. So now the global instance on the server uses that accessToken to make api calls and if a developer does not check if the access token is valid before using the helper data will leak.

1. user with valid accessToken makes a request -> global instance accessToken is set
2. user without token makes a request -> global instance reuses the accessToken from first request

I can think of two possible solutions:

1. throw an error if there is no accessToken
2. unset the accessToken
3. create a new supabase client on every request and store it in locals ([PocketBase](https://github.com/pocketbase/js-sdk#ssr-integration) does it like this)

Option 3 sounds nice, but since we don´t have access to locals in `load` functions this isn´t the best option.

I prefer option 1 as there is no reason to use the helper method if you want to serve public data as you can simply use the global supabaseClient instance. That way you get feedback if you are doing something _wrong_.

kvetoslavnovak:
@david-plugge please, can you check in your updated docs ([Server-side data fetching with RLS ](https://github.com/david-plugge/supabase-auth-helpers/blob/sveltekit-new-api/packages/sveltekit/README.md#server-side-data-fetching-with-rls) and [Protecting API routes](https://github.com/david-plugge/supabase-auth-helpers/blob/sveltekit-new-api/packages/sveltekit/README.md#protecting-api-routes))  whether your imports are right?
Especially importing  ```import { supabaseServerClient, withApiAuth} from '@supabase/auth-helpers-sveltekit';``` but using  ```getSupabaseClient``` and ```withAuth``` later on.

Thank you very much.

david-plugge:
I _think_ this one is ready now. I wonder if supabase-v2 support should also be added?

esojrafael:
Thanks for this @david-plugge hopefully they merge it soon so I can finally upgrade/continue my project.

Torres1080:
Would be pretty nice to have this merged soon 👍

david-plugge:
Alright, ready for review 👍

silentworks:
> I _think_ this one is ready now. I wonder if supabase-v2 support should also be added?

I think v2 support should be kept as a separate PR. I'll review this over this weekend and get it merged in if there are no issues.

silentworks:
Thanks for this excellent work @david-plugge.

kvetoslavnovak:
> Thanks for this @david-plugge hopefully they merge it soon so I can finally upgrade/continue my project.

Same here

jamalmoir:
Thanks @david-plugge you are amazing.

FYI all, I brought this up on Twitter and they have confirmed they are looking at it, hopefully we should be able to use it soon 👍

mitjakukovec:
> I am waiting for this to be merged. Many of my projects are using auth-helpers, so I am unable to update them to the latest svetekit.
> 
> **EDIT** Since this is taking a long time for review I have forked and released it under npm to be able to use it if anyone is interested here's how to install it:
> 
> `npm i -D @supabase/auth-helpers-sveltekit@npm:@ak4zh/auth-helpers-sveltekit@0.6.11`
> 
> This will install the version I released and set an alias `@supabase/auth-helpers-sveltekit` so you can continue to import components from `@supabase/auth-helpers-sveltekit` and when this get's merged just remove mine and install the official one.
> 
> ```
> npm remove @supabase/auth-helpers-sveltekit
> npm i -D @supabase/auth-helpers-sveltekit
> ```

Thank you @ak4zh for your effort in releasing this on npm, even if this is just a temporary solution.

I am wondering what is the reason for [jose](https://www.npmjs.com/package/jose) dependency that was introduced here?
https://www.npmjs.com/package/@ak4zh/auth-helpers-sveltekit?activeTab=dependencies

david-plugge:
> I am wondering what is the reason for [jose](https://www.npmjs.com/package/jose) dependency

It's also used in the [shared](https://www.npmjs.com/package/@supabase/auth-helpers-shared) package to decode the jwt token on the server side to get the expiration timestamp. I was thinking about removing it as a depency and create a decode function since it's kinda heavy with 500kb (tbh. It's not that big) compared to 30kb. However, it's a server only dependency and does not get send to the client. Thoughts?

david-plugge:
@silentworks any updates? :)

mitjakukovec:
> @silentworks any updates? :)

I will give it a try and report back to you on how it goes.

silentworks:
> @silentworks any updates? :)

I've had a look, the magic link example doesn't seem to persist the login. I hit the refresh button in the browser and I'm no longer logged in but I can see that the cookies are still there. Not sure what is going wrong here.

packages/sveltekit/src/server/handlers/auth.ts
```diff
@@ -0,0 +1,8 @@
+import type { Handle } from '@sveltejs/kit';
+import { sequence } from '@sveltejs/kit/hooks';
+import { callback } from './callback';
+import { session } from './session';
+
+export function auth(): Handle {
+  return sequence(callback(), session());
```
silentworks, line 7: I'm not sure if you can wrap a sequence in a sequence with SvelteKit. The reason I'm saying this is that if someone was to use this function, they wouldn't be able to add any additional function to the sequence if they can't wrap a sequence in a sequence.

packages/sveltekit/src/helper.ts
```diff
@@ -0,0 +1,68 @@
...
+  supabaseClient.auth.setAuth(access_token);
+  return supabaseClient;
+}
+
+export function withAuth<E extends any, T extends any>(
+  cb: (event: E & ExtendedEvent) => T | Promise<T>
+): (event: E) => Promise<T> {
+  async function handle(event: E) {
+    let session: SupabaseSession = null;
+    {
```
silentworks, line 37: Probably best to comment how this works as others who are looking to contribute and not very familiar with JavaScript can get tripped up by the scoped evaluations inside of the brackets.

packages/sveltekit/src/helper.ts
```diff
@@ -0,0 +1,68 @@
...
+      // Load
+      const ev = event as unknown as LoadEvent;
+      if (typeof ev.parent === 'function') {
+        const parentData = (await ev.parent()) as App.PageData;
+        session = getClientConfig().getSessionFromPageData(parentData);
+      }
+    }
+
+    const ev = new Proxy(event as E & ExtendedEvent, {
+      get(target, p, receiver) {
```
silentworks, line 54: Can the `p`variable be renamed for better readability and understanding of what `p` represents?

silentworks:
> > I am wondering what is the reason for [jose](https://www.npmjs.com/package/jose) dependency
> 
> It's also used in the [shared](https://www.npmjs.com/package/@supabase/auth-helpers-shared) package to decode the jwt token on the server side to get the expiration timestamp. I was thinking about removing it as a dependency and create a decode function since it's kinda heavy with 500kb compared to 30kb for the helpers. However, it's a server only dependency and does not get send to the client. Thoughts?

We had our own written before and it wasn't able to cover all cloud providers runtime hence why we chose to use `jose` as it works with all the cloud providers and as you said its a server only dependency so not much of a big deal.

packages/sveltekit/src/server/handlers/auth.ts
david-plugge: That shouldn´t be a problem as sequence is basically just a handle function. But i was thinking about removing `auth` anyway since it´s simple to just put `callback` and `session` in there. And some people might prefer not using the callback handle but rather an endpoint.

david-plugge:
> We had our own written before and it wasn't able to cover all cloud providers runtime hence why we chose to use `jose` as it works with all the cloud providers and as you said its a server only dependency so not much of a big deal.

Alright i will revert the commit.

david-plugge:
One question, i removed the shared helpers dependency since the cookies are now set using sveltekit´s cookie api. But the constants, errors, etc. can still be used.
Do you think it should be added for errors, constants etc.?

david-plugge:
> I've had a look, the magic link example doesn't seem to persist the login. I hit the refresh button in the browser and I'm no longer logged in but I can see that the cookies are still there. Not sure what is going wrong here.

It works for me, i´ll take a closer look at it tomorrow.

PS: 
Do you have any plans on adding tests for the helpers?

packages/sveltekit/src/server/handlers/auth.ts
silentworks: I can see that both `callback()` and `session()` are exported too, so we could keep this nice all in one handler along with the individual ones. Documenting them is what will be key to how developers will use them.

silentworks:
> One question, i removed the shared helpers dependency since the cookies are now set using sveltekit´s cookie api. But the constants, errors, etc. can still be used. Do you think it should be added back in?

Yeah this should be added back in.

silentworks:
> > I've had a look, the magic link example doesn't seem to persist the login. I hit the refresh button in the browser and I'm no longer logged in but I can see that the cookies are still there. Not sure what is going wrong here.
> 
> It works for me, i´ll take a closer look at it tomorrow.
> 
> PS: Do you have any plans on adding tests for the helpers?

Yeah there are plans for adding test, just haven't gotten around doing them as yet. Likely to use vitest as the test runner.

david-plugge:
Alright, i read through the `decodeJWT` method in the jose package and it´s almost the same. I think we should drop jose and use a custom decode function in the shared package.


[jose#decodeJWT](https://github.com/panva/jose/blob/main/src/util/decode_jwt.ts)
```ts
import { JWTPayload } from '../types';
import { JWTInvalid, JWTPayloadFailed } from './errors';

const decoder = new TextDecoder();

export function jwtDecoder(jwt: string): JWTPayload {
  const { 1: payload, length } = jwt.split('.');

  if (length !== 3) {
    throw new JWTInvalid();
  }

  try {
    const decoded = Buffer.from(payload, 'base64');
    const result = JSON.parse(decoder.decode(decoded));

    return result;
  } catch (err) {
    throw new JWTPayloadFailed();
  }
}
```

david-plugge:
> its a server only dependency so not much of a big deal

I´m not sure if the package was actually server only. We only rely on tree shaking to remove it as a dep in client side since it´s included in the shared package, which was also loaded on the client.

```js
// packages/shared/dist/index.js
var import_jose = require("jose");
```

david-plugge:
Alright, this looks a lot cleaner to me

silentworks:
> Alright, this looks a lot cleaner to me
> 
> Edit:
> 
> For now i´ve split up server and client config, but since there is no private data passed to the setup method i think they can be combined.
> 
> What do you think?

I agree, combining them will make things much easier too.

silentworks (approved):
LGTM minus the merge conflicts below and combining the setup methods.

david-plugge:
> merge conflicts below

Not sure how to fix them. Can you do it or explain to me how this can be done?

Drewsapple:
If you're still trying to figure out how to resolve the merge conflicts, checkout your `sveltekit-new-api` branch and `git merge {supabase-remote}/main`.

Now that you have the conflicts on your working copy, run `git rm -r packages/svelte` and `git checkout --ours pnpm-lock.json && git add pnpm-lock.json`.

Commit this to your `sveltekit-new-api` branch and it should be good to merge.

david-plugge:
Thanks @Drewsapple! This was the first time i´ve done this, completely forgot that i have to pull from the remote first.

silentworks (approved):

yzn-h:
nice work gays, i really need this to be merged 😅

maxkozlowski:
Great stuff guys, really appreciated all the work that has gone here!
Seconding @yzn-h - if you could merge it, that would help me a lot! 🙏

david-plugge:
> I've forked and [published the new version via npm](https://www.npmjs.com/package/@maxkozlowski/auth-helpers-sveltekit), but getting an error when trying to install.
> `npm ERR! Unsupported URL Type "workspace:": workspace:*`

Have you published the package with npm or pnpm? I think npm doesn´t replace `workspace:*` with the actual version.

And you didn´t publish the auth-helpers-shared package right?

```bash
pnpm i @maxkozlowski/auth-helpers-sveltekit
> ERR_PNPM_NO_MATCHING_VERSION_INSIDE_WORKSPACE  In : No matching version found for @supabase/auth-helpers-shared@* inside the workspace
```

maxkozlowski:
*Have you published the package with npm or pnpm?*
I published it with npm.

*And you didn´t publish the auth-helpers-shared package right?*
That's right, I didn't publish `auth-helpers-shared`.

Happy to try to publish both libraries with pnpm instead if that should work?

david-plugge:
> Happy to try to publish both libraries with pnpm instead if that should work?

It should, if not there is a bug. Make sure to change the package name of the shared helpers in the dependencies of the sveltekit helpers too

silentworks:
Thanks for this @david-plugge, I will get a release out during this week after writing up the migration docs.

maxkozlowski:
Thank you for merging the change and releasing so quickly! I've downloaded the library but I'm getting an error with one of imports.
Does anyone experience the same? Not sure how to fix this :thinking:
`import { auth } from '@supabase/auth-helpers-sveltekit/server';`
_Cannot find module '@supabase/auth-helpers-sveltekit/server' or its corresponding type declarations._

david-plugge:
Yeah that´s a known issue. If you have multiple entry points it´s hard to get ts intellisense working if they are not located in the packages root folder. I´ve setup `publishConfig` in the package.json to publish the dist folder only. It seems like @silentworks decided to remove that. To get it working like this (publishing the whole package, not just dist) we need to add this to the package.json:

```json
"typesVersions": {
  "*": {
    "server": ["./dist/server/index.d.ts"],
    "./": ["./dist/index.d.ts"]
  }
}
```

maxkozlowski:
Thank you David! Does that mean that we'll need to wait for the new release now?

mitjakukovec:
@silentworks, was `publishConfig` in the package.json removed intentionally or if it will be re-introduced? Thanks!

jamalmoir:
I'm getting ERR_MODULE_NOT_FOUND due to imports not having the extensions specified eg. `import {...} from './utils/cookie'` rather than `import {...} from './utils/cookie.js'`

How are people getting this to work?

david-plugge:
> Does that mean that we'll need to wait for the new release now?

In the meantime you can use this alias in your svelte.config.js

```ts
alias: {
	'@supabase/auth-helpers-sveltekit/server': './node_modules/@supabase/auth-helpers-sveltekit/dist/server'
}
```

But i think there are a few more issues that need to be addressed. It seems like `$app/*` is not accessible in a npm package while it works locally in the monorepo.

david-plugge:
> I'm getting ERR_MODULE_NOT_FOUND due to imports not having the extensions specified eg. `import {...} from './utils/cookie'` rather than `import {...} from './utils/cookie.js'`
> 
> How are people getting this to work?

Before the update there was a script that adds `.js` to every imported file to solve that issue. I removed it since it shouldn´t be an issue if the files are located in the packages root. I´ll add that script back in.

jamalmoir:
> > I'm getting ERR_MODULE_NOT_FOUND due to imports not having the extensions specified eg. `import {...} from './utils/cookie'` rather than `import {...} from './utils/cookie.js'`
> > How are people getting this to work?
> 
> Before the update there was a script that adds `.js` to every imported file to solve that issue. I removed it since it shouldn´t be an issue if the files are located in the packages root. I´ll add that script back in.

So rather than it living in node_modules, it should live in your project root?

david-plugge:
> So rather than it living in node_modules, it should live in your project root?

No i mean in the folder `node_modules/@supabase/auth-helpers-sveltekit`. Currently they live in `node_modules/@supabase/auth-helpers-sveltekit/dist`. But i´m not 100% confident about that.

jamalmoir:
> > So rather than it living in node_modules, it should live in your project root?
> 
> No i mean in the folder `node_modules/@supabase/auth-helpers-sveltekit`. Currently they live in `node_modules/@supabase/auth-helpers-sveltekit/dist`. But i´m not 100% confident about that.

Ah, OK I see. Got you, thanks

maxkozlowski:
Has anyone got the new release to work in the end? I am unable to I'm afraid.

It would be fantastic to either get clear instructions as to how to get it to work or ETA when we might get a release that works.

jamalmoir:
@maxkozlowski  https://github.com/supabase/auth-helpers/pull/267

maxkozlowski:
Looks like the new release works well 👏 
The only thing I'm struggling with now is log in. I've followed [the instructions](https://github.com/david-plugge/supabase-auth-helpers/blob/sveltekit-new-api/packages/sveltekit/README.md) and used `saveSession` but getting an error `No action with name 'default' found`.
Has anyone come across this?

ak4zh:
@maxkozlowski You need to create named action. 

`async signin({ request, cookies, url }) `

to

`signin: async ({ request, cookies, url }) `

and in your form use `action=“?/signin”`

david-plugge:
Take a look at this example https://github.com/supabase/auth-helpers/blob/main/examples/sveltekit-email-password/src/routes/(app)/%2Bpage.server.ts :)

ak4zh:
@david-plugge [Here](https://github.com/david-plugge/supabase-auth-helpers/blob/sveltekit-new-api/packages/sveltekit/README.md#hooks-setup) It mentions auth callback `api/auth/callback` do we need to create that endpoint?

or it’s auto handled if we use the `export const handle = auth();`

maxkozlowski:
Thanks both!

One more question: I'm getting some issues when trying to do password reset. I use `supabaseClient.auth.api.resetPasswordForEmail` with user's email and redirect them upon clicking the link to create the new password.

Even though the user appears to be logged in, when I try to do
```
const { user, error } = await supabaseClient.auth.update({
			password: password,
		})
```

I get an error ` Error: Not logged in.`. Has anyone successfully completed password reset and could help?

Thank you 🙏

mitjakukovec:
I followed all the steps on how to setup [@supabase/auth-helpers-sveltekit](https://github.com/supabase/auth-helpers/blob/main/packages/sveltekit/README.md) for SvelteKit

Everything seems so trivial, but when I try to sign into my page using an authentication provider like `google` it will not work.

david-plugge:
@mitjakukovec you can´t get the session like that! `signIn({ provider: 'google' })` returns `{ url: string }`. You need to redirect the user to that url. After the user confirms the propmt by google he will be redirected back to your page with a token in the url hash. The supabase client will read that token and sync the session to the server.

So in case you are trying to achieve oauth login on the server, you simply can´t at the moment (maybe possible in the future).

Just call `supabase.auth.signIn({ provider: 'google' });` on the client and you should be good to go. [example](https://github.com/supabase/auth-helpers/blob/d65b8b4ac6bc6d5248328197dd458abd4e7d804d/examples/sveltekit/src/routes/%2Bpage.svelte#L29)

mitjakukovec:
Thank you @david-plugge, this is exactly what I was doing wrong! I was trying to log in on the server and then somehow sync the session back to the client and not vice versa. As you can see I am not a web developer by trade, but C++ guy with a lot of embedded systems experience.

does it make sense to still do the signing call on the server and then redirect the client to the received url?

david-plugge:
> Does it make sense to do the signin call on the server and then redirect the client to the received URL?

Not really, since you need client js anyway to retrieve the token.
For oAuth signin i´d suggest using the client only to keep it simple :)

PR #258, opened 2022-09-21
chore(nextjs, typescript): allow passing GetServerSideProps and NextApiHandler response types

marpme:

chore: allows developers to passthrough concrete types for `withApiAuth/withPageHandler` handler

Currently we aren't enabled to statically type handlers or getServerSideProps functions with specific return type values, which makes typing hard to ensure functional correctness.

## What is the new behavior?

With this change you would be able to pass-through ResponseType for PageHandlers and/or API Handlers.

Small example for `withApiAuth`: 

```ts
export default withApiAuth<{ text: string }>((req, res) => {
  return res.status(404).send({ text: undefined }) // this will fail
})
```

Small example for `withPageAuth`: 

```ts
export default withPageAuth<{ text: string }>({
  async getServerSideProps() {
    return {
      props: {
        text: undefined // this will fail.
      }
    }
  }
})
```

### DIFF
--- a/packages/nextjs/src/utils/withApiAuth.ts
+++ b/packages/nextjs/src/utils/withApiAuth.ts
@@ -22,8 +22,8 @@ import getAccessToken from './getAccessToken';
  *
  * @category Server
  */
-export default function withApiAuth(
-  handler: NextApiHandler,
+export default function withApiAuth<ResponseType = any>(
+  handler: NextApiHandler<ResponseType>,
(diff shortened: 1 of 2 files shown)


thorwebdev:
Thanks for this. This has been incorporated in the next release. You can already try it out with installing `npm install @supabase/auth-helpers-nextjs@next`  and you can find the migration guide here: https://github.com/supabase/auth-helpers/blob/next/packages/nextjs/MIGRATION_GUIDE.md#migrating-to-03x

PR #267, opened 2022-09-27
fix sveltekit issues

david-plugge:

- [x] type intellisense for `@supabase/auth-helpers-sveltekit/server`
- [x] add `.js` to fix `ERR_MODULE_NOT_FOUND` (can´t use the script, it throws an error `@sveltejs/kit/hooks not found`)
- [x] pass page store and invalidation to startSupabaseSessionSync as `$app` module is not available in npm packages

## Additional context

While this works i´d rather only publish the dist folder generated by `svelte-package` unless there is a good reason not to as it reduces the packages size and fixes the type intellisense and (i think) the `ERR_MODULE_NOT_FOUND` bugs.


silentworks:
@david-plugge You can reinstate the `fix-esm-import-path` npm package which was there before to fix the `.js` issue so you don't have to remember to type them manually.

silentworks (approved):

silentworks (approved):

david-plugge:
> @david-plugge You can reinstate the `fix-esm-import-path` npm package which was there before to fix the `.js` issue so you don't have to remember to type them manually.

I tried but I get an error

> (can´t use the script, it throws an error @sveltejs/kit/hooks not found)

david-plugge:
And why do you prefer publishing the whole folder instead of just the generated dist?

silentworks:
> And why do you prefer publishing the whole folder instead of just the generated dist?

If you try to build any of the example projects inside the monorep you will notice they fail this is due to the way how sveltekit-package does its bundling. Also there was an issue with changeset publishing the package to npm when using the default setup from the sveltekit-package. I've set it up now so only the dist directory will be published along with the README.md, CHANGELOG.md and package.json.

PR #271, opened 2022-09-28
fix: pass domain & path in logout

kangmingtay:
* Allows one to pass the domain and path in handleLogout to clear the cookies. This is particularly useful if you intend to clear all cookies related to a base domain.
* Note: This is just a patch for the version in `main` - we don't need to handle this in `next` since the logout route is removed

### DIFF
--- a/.changeset/dirty-swans-warn.md
+++ b/.changeset/dirty-swans-warn.md
@@ -1,0 +1,5 @@
+---
+'@supabase/auth-helpers-nextjs': patch
+---
+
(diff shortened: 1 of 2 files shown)


thorwebdev:
@kangmingtay have you seen folks complain about this? If not I'd hold off on adding more changes to `main` and rather focus on getting out `next`. Wdyt?

alaister (approved):
lgtm!

thorwebdev (approved):
LGTM, thanks

PR #307, opened 2022-10-13
force session refresh

david-plugge:

Force a session refresh when we are within the specified margin

### DIFF
--- a/packages/sveltekit/src/utils/getServerSession.ts
+++ b/packages/sveltekit/src/utils/getServerSession.ts
@@ -15,6 +15,7 @@ export async function getServerSession(
     session?.expires_at &&
     session.expires_at + expiry_margin <= Date.now() / 1000
   ) {
+    session.access_token = '';
     const refreshed = await supabase.auth.setSession(session);
 
     session = refreshed.data.session;


j4w8n:
I'm not sure who the decision-makers are for all of the supabase/go-true/auth-helpers stuff, but per [this comment](https://github.com/supabase/gotrue-js/issues/490#issuecomment-1279932972), allowing the session `access_token` to be an empty string for `setSession()` was unintended. I'm currently using this work-around too, to force a refresh even outside the `expiry_margin`; just passing along what I've been told.

j4w8n:
As of supabase-js v2.0.4, the `refreshSession()` method is available. It'll refresh tokens, and return the new session, no matter when the access_token expires - and even if it has expired.

If run on the client-side, without passing in anything, it'll grab session info from `getSession()` and use that info to refresh.

When calling it from the server-side, you'll want to pass in a session object with the refresh token.
`supabase.auth.refreshSession({ refresh_token: 'token' })`

david-plugge:
Thanks @j4w8n, i´ll close this pr in favor of #327

Issue #289, opened 2022-10-10
Update sveltekit auth helpers to v2


Nisthar:
Please update the sveltekit auth helpers to v2.

MentalGear:
The have just done it, but is it working for you @Nisthar ? I get an `Cannot destructure property 'supabaseUrl' of 'getConfig(...)' as it is undefined.` error.

f-gueguen:
@MentalGear I think in `lib/db.js` you are importing the `supabase-js` createClient function instead of from `auth-helpers-sveltekit`

MentalGear:
Indeed, thanks again @f-gueguen !

CanRau:
Yaaasss nice, also successfully upgraded to v2 today following the readme of SvelteKit auth-helpers

braden-w:
Have been unable to pass `Database` via `createClient<Database>(supabaseUrl, supabaseAnonKey)` like in v2, @CanRau  could you elaborate on how you were able to upgrade to v2?

So far, in order to use types on selects and updates, I've done the following in my `lib/db.ts` : 

```ts
// From https://github.com/supabase/auth-helpers/blob/main/packages/sveltekit/README.md 
import { createClient } from '@supabase/auth-helpers-sveltekit';
import { createClient as createFetchClient } from '@supabase/supabase-js';
import { env } from '$env/dynamic/public'
import type { Database } from '$lib/types/DatabaseDefinitions';

const supabaseUrl = env.PUBLIC_SUPABASE_URL;
const supabaseAnonKey = env.PUBLIC_SUPABASE_ANON_KEY;

export const supabaseClient = createClient(supabaseUrl, supabaseAnonKey);
export const supabaseFetchClient = createFetchClient<Database>(supabaseUrl, supabaseAnonKey);
```

and use `supabaseClient` for most operations, and `supabaseFetchClient` for CRUD operations when I want types.

CanRau:
I'm still not using types that's how 🙈
First time I tried to run locally with the CLI I ran into some issues and didn't try again so far 😅

PR #326, opened 2022-10-21
Specify DatabaseDefinitions origins for Sveltekit (Get types working)

braden-w:
Clarifies confusion about how DatabaseDefinitions was created using Supabase CLI. This clears some potential confusion and clarifies #289, where the user should have the import point to their types file.

### DIFF
--- a/packages/sveltekit/README.md
+++ b/packages/sveltekit/README.md
@@ -119,7 +119,7 @@ This results in less server calls as the client manages the session on it´s own
 
 ### Typings with v2 from Supabase CLI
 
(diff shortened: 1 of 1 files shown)


packages/sveltekit/README.md
```diff
@@ -117,9 +117,9 @@ export const load: LayoutLoad = async (event) => {
 
 This results in less server calls as the client manages the session on it´s own.
 
-### Typings
+### Typings with v2 from Supabase CLI
 
-In order to get the most out of TypeScript and it´s intellisense, you should import our types into the `app.d.ts` type definition file that comes with your SvelteKit project.
+In order to get the most out of TypeScript and it´s intellisense, you should import our types into the `app.d.ts` type definition file that comes with your SvelteKit project, where `import('./DatabaseDefinitions')` points to the generated types file outlined in [v2 docs here](https://supabase.com/docs/reference/javascript/release-notes#typescript-support) after you have logged in, linked, and generated types through the Supabase CLI.
```
david-plugge, line 122: We should get rid of `import our types` since we don´t export any types for that use case.
Something like `import the generated Database types`

silentworks (approved):

PR #327, opened 2022-10-23
update supabase js

david-plugge:

Update to supabase-js 2.0.4

We are currently missusing `setSession(session)` to refresh the session in sveltekit.

## What is the new behavior?

Use the new `refreshSession()` method

### DIFF
--- a/packages/sveltekit/package.json
+++ b/packages/sveltekit/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@supabase/auth-helpers-sveltekit",
-  "version": "0.8.3",
+  "version": "0.8.2",
   "description": "A collection of framework specific Auth utilities for working with Supabase.",
   "type": "module",
   "main": "./dist/index.js",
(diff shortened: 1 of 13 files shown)


david-plugge:
> This seems a duplicate of #331

No, this PR makes use of the new method `refreshSession` as described above

silentworks:
@david-plugge do we need this to be a new release?

wiesson:
Does it make sense to split the PR in 

1. Update supabase-js to most recent version
2. Switch to refreshSession

So it's more clear, what is going on

packages/sveltekit/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@supabase/auth-helpers-sveltekit",
-  "version": "0.8.2",
+  "version": "0.8.3",
```
silentworks, line 3: This shouldn't be manually adjusted. `changeset` manages this for us.

silentworks:
> Does it make sense to split the PR in
> 
>     1. Update supabase-js to most recent version
> 
>     2. Switch to refreshSession
> 
> 
> So it's more clear, what is going on

Its only a small change but we can write this in the changeset.

silentworks (approved):

Issue #330, opened 2022-10-24
Export error in latest sveltekit-auth-helper


ak4zh:
```bash
Error: Missing "./server" export in "@supabase/auth-helpers-sveltekit" package
    at bail (file:///Users/ak4zh/CodeProjects/sveltekit-saas-starter/node_modules/.pnpm/vite@3.1.8/node_modules/vite/dist/node/chunks/dep-4da11a5e.js:33715:8)
    at resolve (file:///Users/ak4zh/CodeProjects/sveltekit-saas-starter/node_modules/.pnpm/vite@3.1.8/node_modules/vite/dist/node/chunks/dep-4da11a5e.js:33792:10)
    at resolveExports (file:///Users/ak4zh/CodeProjects/sveltekit-saas-starter/node_modules/.pnpm/vite@3.1.8/node_modules/vite/dist/node/chunks/dep-4da11a5e.js:34853:12)
    at resolveDeepImport (file:///Users/ak4zh/CodeProjects/sveltekit-saas-starter/node_modules/.pnpm/vite@3.1.8/node_modules/vite/dist/node/chunks/dep-4da11a5e.js:34871:31)
    at tryNodeResolve (file:///Users/ak4zh/CodeProjects/sveltekit-saas-starter/node_modules/.pnpm/vite@3.1.8/node_modules/vite/dist/node/chunks/dep-4da11a5e.js:34584:20)
    at Context.resolveId (file:///Users/ak4zh/CodeProjects/sveltekit-saas-starter/node_modules/.pnpm/vite@3.1.8/node_modules/vite/dist/node/chunks/dep-4da11a5e.js:34357:28)
    at Object.resolveId (file:///Users/ak4zh/CodeProjects/sveltekit-saas-starter/node_modules/.pnpm/vite@3.1.8/node_modules/vite/dist/node/chunks/dep-4da11a5e.js:41039:46)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async TransformContext.resolve (file:///Users/ak4zh/CodeProjects/sveltekit-saas-starter/node_modules/.pnpm/vite@3.1.8/node_modules/vite/dist/node/chunks/dep-4da11a5e.js:40808:23)
    at async normalizeUrl (file:///Users/ak4zh/CodeProjects/sveltekit-saas-starter/node_modules/.pnpm/vite@3.1.8/node_modules/vite/dist/node/chunks/dep-4da11a5e.js:37589:34)
```

```
[...]

PR #331, opened 2022-10-24
allow supabase patch updates

ak4zh:
allow installing newer patch updates

Does not allow to install newer version of supabase `2.0.4`


KevM:
Just did a `npm update` today and get warning it seems this PR would fix. 
image

david-plugge:
It should be enough to modify the peerDependency. I think we should prefer specific versions to build the package.

david-plugge:
> Any idea what’s causing this issue #330

The old version used two different exports. You probably have an import `@supabase/auth-helpers-sveltekit/server` somewhere in your code that should be updated

silentworks:
Pr #327 is addressing this issue.

PR #333, opened 2022-10-25
Feat: Add Remix auth helpers

dijonmusters:
remix

feature

This adds new Auth Helpers for the Remix framework, as well as an example application

No auth helpers for Remix

## What is the new behavior?

Auth helpers for Remix


.changeset/brown-students-act.md
```diff
@@ -0,0 +1,6 @@
+---
+'@example/remix': minor
```
thorwebdev, line 2: IIRC we don't version examples. I think can just remove.
```suggestion
```

examples/remix/app/routes/conditional-ssr.tsx
```diff
@@ -0,0 +1,46 @@
...
+  request
+}: {
+  request: Request;
+}) => {
+  const response = new Response();
+  const supabaseClient = getSupabase<Database>({ request, response });
+
+  const {
+    data: { session }
+  } = await supabaseClient.auth.getSession();
```
thorwebdev, line 18: can simplify:

```js
const {
    data: { user }
  } = await supabaseClient.auth.getUser();
```

packages/remix/README.md
```diff
@@ -0,0 +1,428 @@
...
+
+Using [yarn](https://yarnpkg.com/):
+
+```sh
+yarn add @supabase/auth-helpers-remix
+```
+
+This library supports the following tooling versions:
+
+- Node.js: `^10.13.0 || >=12.0.0`
```
thorwebdev, line 21: check if remix has this limitation

thorwebdev (approved):
LGTM, thanks 🥳

danielvanc:
👏🏼 Looks great!

packages/remix/package.json
```diff
@@ -0,0 +1,48 @@
+{
+  "name": "@supabase/auth-helpers-remix",
+  "version": "0.4.0",
```
silentworks, line 3: You might want to reset this version number to 0.0.0 and make the changeset a minor release which will bump it to 0.1.0 in the GitHub action process.

examples/remix/app/routes/conditional-ssr.tsx
silentworks: It's probably best to keep it as `getSession`, since this function will refresh the session if necessary or get the session from inMemory while `getUser` won't do this.

dijonmusters: Good call!

packages/remix/package.json
dijonmusters: Good call!

silentworks (approved):

dijonmusters:
After some discussion with @alaister about code splitting, I have decided to refactor this as two separate functions:

1. `getServerClient` to be called in `loader` and `action` functions, and requires a Request and Response
2. `getBrowserClient` to be called in `useEffect` and does not require a Request or Response

I think this is also more clear than one isomorphic magic function that sometimes is called with a request and response and sometimes not

PR #338, opened 2022-10-25
Ts improvements

ak4zh:
Docs update
Added some missing types in the docs.

### DIFF
--- a/packages/sveltekit/README.md
+++ b/packages/sveltekit/README.md
@@ -106,10 +106,10 @@ In addition you can create a layout load function if you are using `invalidate('
 
 ```ts
 // src/routes/+layout.ts
-import type { LayoutLoad } from './$types';
+import type { LayoutLoad, LayoutLoadEvent } from './$types';
 import { getSupabase } from '@supabase/auth-helpers-sveltekit';
 
-export const load: LayoutLoad = async (event) => {
+export const load: LayoutLoad = async (event: LayoutLoadEvent) => {
   const { session } = await getSupabase(event);
   return { session };
 };
--- a/packages/sveltekit/README.md
+++ b/packages/sveltekit/README.md
@@ -92,10 +92,10 @@ In order to make the session available to the UI (pages, layouts) we need to pas
 
 ```ts
 // src/routes/+layout.server.ts
-import type { LayoutServerLoad } from './$types';
+import type { LayoutServerLoad, LayoutServerLoadEvent } from './$types';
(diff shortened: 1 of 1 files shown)


david-plugge:
I don´t think there´s any benefit? The types from `./$types` are used to get typescript intellisense working. By typing the event we loose all benefits of sveltekits type generation.

Why do you think this should be added, is intellisense not working for you?

ak4zh:
@david-plugge I am new to typescript.
I just added these because typescript was not happy until I added these in my project. (files were marked red)
All my files were red without these, so I thought may be it will be a good idea to have this in the docs itself so next time I copy paste the code to implement auth-helpers I don’t have to apply these fixes to remove the typescript warnings.

david-plugge:
@ak4zh It seems like there is something wrong with your typescript setup. Can you share a repo?

ak4zh:
@david-plugge https://github.com/ak4zh/sveltekit-saas-starter

While you are giving it a look can you also see if you know why the auth-helps give me another error only in production.
Here’s the deployed version of same repo https://sveltekit-saas-starter.vercel.app/ check the console it will have an error that supabaseUrl is required, while same code works fine in dev mode.
This has to do something with the new version of auth-helpers

david-plugge:
> While you are giving it a look can you also see if you know why the auth-helps give me another error only in production.
> Here’s the deployed version of same repo [sveltekit-saas-starter.vercel.app](https://sveltekit-saas-starter.vercel.app/) check the console it will have an error that supabaseUrl is required, while same code works fine in dev mode.
> This has to do something with the new version of auth-helpers

That´s because you are using dynamic env variables. In that case you need to set the env variables in the vercel deployment.
You can use static env if you want to replace the variables during build time.
```ts
import { env } from '$env/static/public';
```
https://github.com/ak4zh/sveltekit-saas-starter/blob/bc1d74a61ebebc0176b35089a081838d6b795517/src/lib/db.ts#L2

david-plugge:
You are missing a tsconfig.json in your project. Rename your jsconfig to tsconfig.

ak4zh:
@david-plugge I do have env variables setup from vercel dashboard.
Not really sure about the difference between dynamic and static.

Though the error says missing `supabaseUrl` but if you inspect the page source you will find all env are available inside.
I will try with static

david-plugge:
@ak4zh are you still running into the type issue?

ak4zh:
No it seems everything is fixed now.
Thanks for the assistance.

PR #350, opened 2022-10-31
chore: reduce cookie size

thorwebdev:
Strip down the cookie to only store the tokens and construct the session by decoding the `access_token`.

In a next step we should look into splitting the cookie up into chunks since the `access_token` itself could exceed 4kb.

### DIFF
--- a/packages/shared/src/utils/cookies.ts
+++ b/packages/shared/src/utils/cookies.ts
@@ -45,6 +45,24 @@ export function isSecureEnvironment(headerHost?: string | string[]) {
   return true;
 }
 
+const decodeBase64URL = (value: string): string => {
+  try {
+    // atob is present in all browsers and nodejs >= 16
+    // but if it is not it will throw a ReferenceError in which case we can try to use Buffer
+    // replace are here to convert the Base64-URL into Base64 which is what atob supports
+    // replace with //g regex acts like replaceAll
+    return atob(value.replace(/[-]/g, '+').replace(/[_]/g, '/'));
+  } catch (e) {
+    if (e instanceof ReferenceError) {
+      // running on nodejs < 16
+      // Buffer supports Base64-URL transparently
+      return Buffer.from(value, 'base64').toString('utf-8');
+    } else {
+      throw e;
+    }
+  }
+};
+
 export function parseSupabaseCookie(
   str: string | null | undefined
 ): Partial<Session> | null {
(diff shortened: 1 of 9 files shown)


packages/shared/src/utils/cookies.ts
```diff
@@ -43,3 +44,55 @@ export function isSecureEnvironment(headerHost?: string | string[]) {
...
+    session.provider_token,
+    session.provider_refresh_token
+  ]);
+}
+
+function base64urlToString(base64url: string) {
+  if (typeof Buffer !== 'undefined') {
+    return Buffer.from(base64url, 'base64url').toString('utf-8');
+  }
+  return atob(base64url);
```
thorwebdev, line 97: Do we need to transfrom base64 url to base64 like so? https://github.com/supabase/gotrue-js/blob/master/src/lib/helpers.ts#L87

packages/shared/src/utils/cookies.ts
```diff
@@ -43,3 +44,66 @@ export function isSecureEnvironment(headerHost?: string | string[]) {
...
+    // atob is present in all browsers and nodejs >= 16
+    // but if it is not it will throw a ReferenceError in which case we can try to use Buffer
+    // replace are here to convert the Base64-URL into Base64 which is what atob supports
+    // replace with //g regex acts like replaceAll
+    return atob(value.replace(/[-]/g, '+').replace(/[_]/g, '/'));
+  } catch (e) {
+    if (e instanceof ReferenceError) {
+      // running on nodejs < 16
+      // Buffer supports Base64-URL transparently
+      return Buffer.from(value, 'base64').toString('utf-8');
```
silentworks, line 59: We've had issue in the past with Buffer not being available on cloudflare workers If I remember correctly, can we test this out before getting this merged in.

thorwebdev: That's why it falls back to `atob` it's taken from gotrue-js

thorwebdev: https://developers.cloudflare.com/workers/runtime-apis/web-standards/#base64-utility-methods

packages/shared/src/utils/cookies.ts
```diff
@@ -43,3 +44,66 @@ export function isSecureEnvironment(headerHost?: string | string[]) {
...
+      provider_token: session[2],
+      provider_refresh_token: session[3],
+      user: {
+        id: sub,
+        ...user
+      }
+    };
+  } catch (err) {
+    return null;
+  }
```
hf, line 99: This should also be handling the case where the session string has the previous format?

thorwebdev: Good idea, yes.

packages/shared/src/utils/cookies.ts
thorwebdev: Tested and confirmed that it works both in cludflare workers as well as vercel edge functions 👍

sorin89:
@thorwebdev thanks for this fix. I think I figured out why the token size in my case was so large: I am using Google Workspace (G suite) account, and the profile pic url was very long (995 characters), whereas a usual profile pic url is less than 100 characters.

PR #356, opened 2022-11-01
chore: move to client-side auth and useOutletContext.

thorwebdev:
This moves the Auth requests to the client-side with Auth UI and uses `useOutletContext` to hand the browser client through to the child components.

This approach will simplify things and save you money (server function invocations).


dijonmusters:
> This moves the Auth requests to the client-side with Auth UI and uses `useOutletContext` to hand the browser client through to the child components.
> 
> This approach will simplify things and save you money (server function invocations).

I do like saving money!

examples/remix/app/root.tsx
```diff
@@ -150,57 +102,28 @@ export default function App() {
...
+            redirectTo="http://localhost:3004"
+            appearance={{ theme: ThemeSupa }}
+            supabaseClient={supabase}
+            providers={['google', 'github']}
+            socialLayout="horizontal"
+          />
+        )}
         <hr />
-        <Outlet />
+        <Outlet context={context} />
```
dijonmusters, line 125: This is much cleaner, good call 👍

dijonmusters (approved):
LGTM 👍 Thanks Thor!

PR #362, opened 2022-11-02
Environment variables NOT available in static builds

renefournier:
Prior code means variables are not available when statically built and/or deployed to Netlify or Vercel. 

Fixes `npm run build` error that environment variables are not available.

“supabaseUrl is required” error message when deploying statically.

## What is the new behavior?

Error is gone, **supabase** works in `dev` and `production`, server-side and client-side as advertised.

## Additional context


david-plugge:
If we want `static` to be the default for env variables we should update all examples and maybe also the docs.
https://github.com/supabase/auth-helpers/blob/main/packages/sveltekit/README.md#supabaseclient-setup

david-plugge:
We decided to not change the way we load env variables by default. You can always change it to your likings.

PR #380, opened 2022-11-14
Add next.js auth helpers for server components

dijonmusters:
Add an additional auth helper function for `createServerComponentSupabaseClient`. This, paired with a middleware function to refresh the Supabase session, make it possible to use an authenticated Supabase client in a Next.js 13 Server Component - the `app` directory.

The `app` directory is still in beta so is likely to change. This should be seen as an experimental approach to start trying out Server Components with Supabase.

There is no supported way to create an authenticated Supabase client in a server component

https://github.com/supabase/auth-helpers/issues/341

## What is the new behavior?

We have an experimental way to create an authenticated Supabase client in a server component


thorwebdev (approved):
LGTM, thanks.

PR #389, opened 2022-11-24
Qwik adaptor

TheDeveloper:

This PR adds an adaptor for Supabase auth in [Qwik](https://qwik.builder.io/). This makes it possible to use Supabase Auth in Qwik on the Client and in Server Side Rendering (SSR).

Usage instructions and examples are available in the README: https://github.com/TheDeveloper/supabase-auth-helpers/tree/qwik-adaptor/packages/qwik


silentworks:
The Qwik team has created a their own version which they will maintain. https://github.com/BuilderIO/qwik/tree/main/packages/supabase-auth-helpers-qwik

PR #397, opened 2022-12-06
Chore: Update next.js server components example

dijonmusters:

Simplifies the Next.js 13 Server Components example

Please link any relevant issues here.

## What is the new behavior?

- instantiating server client is cleaner
- more clear separation of client and server
- single supbase instance for client-side (shared through context)
- fix useEffect dependencies


thorwebdev (approved):
LGTM, thanks

PR #399, opened 2022-12-07
Chore: Simplify remix example

dijonmusters:

Simplify Remix example and make it as consistent as possible with Next.js Server Components

Remix example is hard to follow

## What is the new behavior?

Remix example covers:

1. password auth
2. GitHub OAuth
3. server-side Supabase client
4. client-side singleton Supabase instance - shared through outlet context
5. 100% TypeScript (server and client)
6. optional session page
7. required session page
8. realtime (merging client and server state)
9. mutations


thorwebdev (approved):
LGTM, thanks

Issue #396, opened 2022-12-03
Accept client options when creating a client using the auth helpers


Murkrage:
# Feature request

## Is your feature request related to a problem? Please describe.
When creating a client using `supabase-js` you have the option to use a [custom fetch implementation](https://supabase.com/docs/reference/javascript/initializing#custom-fetch-implementation). This currently isn't possible when creating either a browser or server client using the auth helpers.

## Describe the solution you'd like

In order to make use of Next 13's caching capabilities, I would like to be able to use their fetch implementation (which extends the native Fetch) with Supabase through the auth helpers. I think the best way to go about this is to allow us to provide the same client options we can provide using `createClient()` directly from the main library.

## Additional context

Info on Next 13's fetch: https://beta.nextjs.org/docs/api-reference/fetch

thorwebdev:
The supabase client uses the native fetch by default. So you don't need to pass it in for Next.js 13. You can see some examples here: https://github.com/supabase/auth-helpers/blob/main/examples/nextjs-server-components/app/optional-session/page.tsx#L7-L8

PR #410, opened 2022-12-30
[Breaking] Allow initial responses from supabase to be reused from the server when doing SSR

### LINKED ISSUE #409: TypedSupabaseClient should be moved directly to `@supabase/auth-helpers-sveltekit`
# Bug report

## Describe the bug

The type `TypedSupabaseClient` is currently in `@supabase/auth-helpers-sveltekit/dist/types`, when imported. 

## To Reproduce

Try to import `TypedSupabaseClient`

## Expected behavior

It should be imported from `@supabase/auth-helpers-sveltekit` instead.

## System information

- OS: Windows
```
		"@supabase/auth-helpers-sveltekit": "0.8.6",
		"@supabase/supabase-js": "2.2.2"
```

zlepper:
This is basically the SvelteKit version of #396, which is why I have not created a new issue. 

Also, this is the first time I'm doing a PR for this project, so please do point out if some things should be done differently. 

One of SvelteKit's features is that if you use the `fetch` function provided in the load functions, you can avoid fetching data the server just fetched when the client loads.

This is a breaking change, requiring a slight change to the `hooks.server.ts/js` file. Specifically:

> In the `src/hooks.server.js` we also need to whitelist some Supabase headers to prevent them from being stripped out by SvelteKit:
>
> If you don't have a custom `handle` hook, then you can add the following:
> ```ts
import { allowSupabaseServerSideRequests } from '@supabase/auth-helpers-sveltekit';

export const handle = allowSupabaseServerSideRequests;
```
> Otherwise, you can add the following to your existing `handle` hook:
> ```ts
import { RequiredWhitelistedHeaders } from '@supabase/auth-helpers-sveltekit';

export async function handle({ event, resolve }) {
  // ... your existing code

  const response = await resolve(event, {
    // This is the important bits.
    filterSerializedResponseHeaders(name: string, _value: string): boolean {
      return RequiredWhitelistedHeaders.includes(name);
    }
  });

  // ... More of your code

  return response;
}
```

I also fixed #409 to enable `TypedSupabaseClient` to be imported directly from `@supabase/auth-helpers-sveltekit` while I was here.

Data is loaded twice, first on the server, then on the client on the initial navigation. 

## What is the new behavior?

Data is only loaded once, server side, on the initial load. Afterwards purely client side loading is used.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,6 @@ docs/
 
 # svelte kit
 .svelte-kit
+
+# Intellij IDEs
+.idea/
--- a/packages/sveltekit/src/hooks.server.ts
+++ b/packages/sveltekit/src/hooks.server.ts
@@ -1,0 +1,16 @@
+import type {Handle} from "@sveltejs/kit";
+
+export const RequiredWhitelistedHeaders: ReadonlyArray<string> = [
+    'content-range'
+]
+
+/**
+ * A default implementation of the `handle` hook, in case you don't need to do anything special.
+ */
+export const allowSupabaseServerSideRequests: Handle = function({event, resolve}) {
+    return resolve(event, {
+        filterSerializedResponseHeaders(name: string, _value: string): boolean {
+            return RequiredWhitelistedHeaders.includes(name);
+        }
+    })
+}
--- a/packages/sveltekit/src/utils/supabase-load.ts
+++ b/packages/sveltekit/src/utils/supabase-load.ts
@@ -1,18 +1,18 @@
 import { getConfig } from '../config';
-import { isBrowser } from '@supabase/auth-helpers-shared';
 import { createClient, type Session } from '@supabase/supabase-js';
 import type { LoadEvent } from '@sveltejs/kit';
 import type { TypedSupabaseClient } from '../types';
 
 export function getLoadSupabaseClient(event: LoadEvent): TypedSupabaseClient {
-  const { supabaseUrl, supabaseKey, options, globalInstance } = getConfig();
-
-  if (isBrowser()) {
-    return globalInstance;
-  }
+  const { supabaseUrl, supabaseKey, options } = getConfig();
 
   return createClient(supabaseUrl, supabaseKey, {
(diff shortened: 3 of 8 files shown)


.gitignore
```diff
@@ -38,4 +38,7 @@ docs/
 .turbo
 
 # svelte kit
-.svelte-kit
\ No newline at end of file
+.svelte-kit
+
+# Intellij IDEs
+.idea/
```
zlepper, line 44: WebStorm tried to add some files to git, added this to prevent that.

packages/sveltekit/src/hooks.server.ts
```diff
@@ -0,0 +1,16 @@
+import type {Handle} from "@sveltejs/kit";
+
+export const RequiredWhitelistedHeaders: ReadonlyArray<string> = [
+    'content-range'
+]
```
zlepper, line 5: Technically this could be done by just having a direct string here, but I'm not sure if Supabase has other headers, so to avoid breaking people again in case this needs to change, I thought I might as well keep it as an array right away.

packages/sveltekit/src/utils/supabase-load.ts
```diff
@@ -1,18 +1,18 @@
...
 import { createClient, type Session } from '@supabase/supabase-js';
 import type { LoadEvent } from '@sveltejs/kit';
 import type { TypedSupabaseClient } from '../types';
 
 export function getLoadSupabaseClient(event: LoadEvent): TypedSupabaseClient {
-  const { supabaseUrl, supabaseKey, options, globalInstance } = getConfig();
-
-  if (isBrowser()) {
-    return globalInstance;
-  }
```
zlepper, line 12: I'm aware this will cause each load function to new up a new client instance, but I think the trade-off is worth it here, considering we are removing requests the client has to do.

david-plugge:
Thanks alot for mentioning this issue! As be are about to release the new version of auth helpers that include this functionality i´m going to close this issue.

PR #411, opened 2022-12-30
Fix getServerSession parameter type

epavanello:
"getServerSession" is mainly used inside a "LayoutServerLoad" function (it expose an event typed "ServerLoadEvent"), but the event object is only propagated to "getRequestSupabaseClient" that accept a generic "RequestEvent". It doesn't work if I need to get the server session from an api (see https://kit.svelte.dev/docs/routing#server) that by "RequestHandler" offer only a generic "RequestEvent" insead of an "ServerLoadEvent".

Bug fix

getServerSession has incompatible event type if used by apis (see https://kit.svelte.dev/docs/routing#server)

## What is the new behavior?

Only a replacement for a more generic type

### DIFF
--- a/packages/sveltekit/src/utils/getServerSession.ts
+++ b/packages/sveltekit/src/utils/getServerSession.ts
@@ -1,8 +1,8 @@
-import type { ServerLoadEvent } from '@sveltejs/kit';
+import type { RequestEvent } from '@sveltejs/kit';
 import { getRequestSupabaseClient } from './supabase-request';
 
 export async function getServerSession(
-  event: ServerLoadEvent,
+  event: RequestEvent,
   expiry_margin = 60
 ) {
   const supabase = getRequestSupabaseClient(event);


david-plugge:
Thanks for pointing this out.

`ServerLoadEvent` is the correct type here as we only use `getServerSession` to pass the session to universal load functions.
It´s a little complicated to explain why we need this helper.

The flow looks like this:

1. `handle` in `hooks.server.js`
2. server only load functions
3. shared load functions

Inside the server load functions we can use `getSupabase` to get the supabase client instance. This instance is able to override the session cookie because we have access to `cookies` inside `ServerLoadEvent`. Same thing on `+server` request handlers.

But when a universal load function runs on the server we don´t have a way to override the session cookie (refresh the session when it´s expired or close to beeing expired).
`getServerSession` returns the session if it is currently valid for more than a set number of seconds and otherwise refreshes the clients session. That way we can make sure the client will never try to refresh the session in a universal load running on the server.

I hope this explains the usage of `getServerSession`.

So to sum it up, this PR should be closed and the docs should make clear why this helper function is needed.

But i didn´t had a lot of time to read up what changed in the supabase client package so we maybe don´t even need to include this helper anymore.

david-plugge:
This is by design, you are meant to use `const { session } = await getSupabase(event)` in all load functions/requestHandlers.
getServerSession should only be used inside the root layout load to propagate the session to shared load functions. We should rename getServerSession to something more verbose

PR #427, opened 2023-01-17
feat: allow passing client options

david-plugge:

Allow passing of client options for nextjs and remix https://github.com/supabase/auth-helpers/issues/420

Client options can´t be passed.

## What is the new behavior?

client options can be passed:

```ts
import { createBrowserSupabaseClient } from '@supabase/auth-helpers-nextjs';

const supabaseClient = createBrowserSupabaseClient({
	options: {
		...
	}
});
```

The auth key is omitted as it is set internally by the helper

### DIFF
--- a/packages/sveltekit/src/createClient.ts
+++ b/packages/sveltekit/src/createClient.ts
@@ -1,7 +1,7 @@
 import {
   createBrowserSupabaseClient,
   type CookieOptions,
-  type SupabaseClientOptions
+  type SupabaseClientOptionsWithoutAuth
 } from '@supabase/auth-helpers-shared';
 import { setConfig } from './config';
 import { PKG_NAME, PKG_VERSION } from './constants';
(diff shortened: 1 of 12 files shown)


packages/sveltekit/src/createClient.ts
```diff
@@ -11,7 +10,7 @@ import type { TypedSupabaseClient } from './types';
 export function createClient(
   supabaseUrl: string,
   supabaseKey: string,
-  options?: SupabaseClientOptionsWithoutAuth<App.Supabase['SchemaName']>,
```
thorwebdev, line 14: Actually, I'd say the sveltekit option implementation is exactly what we want to use across the board. Sorry if that wasn't clear.

david-plugge: It should be the same for all, the SupabaseClientOptionsWithoutAuth type was unnecessary as the type `SupabaseClientOptions` exported from the shared package omits the auth key
https://github.com/supabase/auth-helpers/blob/7ddaa6a1dbb9262ea15620d661b6d1691a23623f/packages/shared/src/types.ts#L9

thorwebdev: ah yes, perfect, I missed that it was coming from shared, sweet 👍

david-plugge: Maybe we should also rename the type to reduce confusion

thorwebdev: Yah, I think that might be good

thorwebdev (approved):
LGTM, thanks

thorwebdev:
@david-plugge if this is ready, can you run changeset for it? https://github.com/supabase/auth-helpers/blob/main/development.md#preparing-a-release

PR #428, opened 2023-01-18
Allow passing client options

david-plugge:

Allow passing of client options for nextjs and remix https://github.com/supabase/auth-helpers/issues/420

Client options can´t be passed.

## What is the new behavior?

client options can be passed:

```ts
import { createBrowserSupabaseClient } from '@supabase/auth-helpers-nextjs';

const supabaseClient = createBrowserSupabaseClient({
	options: {
		...
	}
});
```

The auth key is omitted as it is set internally by the helper


david-plugge:
Recreated due to conflicting files

silentworks (approved):

PR #430, opened 2023-01-18
feat: next allow overwriting url and key

### LINKED ISSUE #384: [Next.js] Allow overwriting of client URL and key.
When using `createMiddlewareSupabaseClient` it is not possible to set api url and key. In the source code it seems like it takes it directly from the environment variables, which I think is fine for fallback.
However I would like to define my environment variables as I like and not be dependent on how library maintainers require my envrionment to be configured.

I suggest adding one or two paramenters more to the function such that the user can define api url and key.

Thanks in advance

Related code:
https://github.com/supabase/auth-helpers/blob/2be3f1001c5770db591d1d84cec02f4648ecfd75/packages/nextjs/src/index.ts#L123-L124

david-plugge:

Allow overwriting of client URL and key in Next.js helpers
fixes #384

Values must be set from env variables

## What is the new behavior?

Values can be set in create*SupabaseClient functions

```ts
createBrowserSupabaseClient({
	supabaseUrl: 'supabase-url',
	supabaseKey: 'supabase-anon-key',
})
```

Defaults to the env variables so this isn´t a breaking change

## Additional context

Added options for `createServerComponentSupabaseClient` as that was missed in #428

### DIFF
--- a/packages/nextjs/src/index.ts
+++ b/packages/nextjs/src/index.ts
@@ -31,24 +31,25 @@ export function createBrowserSupabaseClient<
     ? 'public'
     : string & keyof Database
 >({
+  supabaseUrl = process.env.NEXT_PUBLIC_SUPABASE_URL,
+  supabaseKey = process.env.NEXT_PUBLIC_SUPABASE_ANON_KEY,
   options,
   cookieOptions
 }: {
+  supabaseUrl?: string;
+  supabaseKey?: string;
   options?: SupabaseClientOptionsWithoutAuth<SchemaName>;
   cookieOptions?: CookieOptions;
 } = {}) {
-  if (
-    !process.env.NEXT_PUBLIC_SUPABASE_URL ||
-    !process.env.NEXT_PUBLIC_SUPABASE_ANON_KEY
-  ) {
+  if (!supabaseUrl || !supabaseKey) {
     throw new Error(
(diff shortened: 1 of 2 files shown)


logemann:
really looking forward to this PR. Changes do exactly what i am currently looking for.

packages/nextjs/src/index.ts
```diff
@@ -31,24 +31,25 @@ export function createBrowserSupabaseClient<
     ? 'public'
     : string & keyof Database
 >({
+  supabaseUrl = process.env.NEXT_PUBLIC_SUPABASE_URL,
+  supabaseKey = process.env.NEXT_PUBLIC_SUPABASE_ANON_KEY,
   options,
   cookieOptions
 }: {
+  supabaseUrl?: string;
+  supabaseKey?: string;
```
logemann, line 40: it would have been nice to have an additional: useServiceKey?: boolean with auto-lookup like the other two env vars. This would be a super easy way to just tell the client that i want to work with the service key instead of anon key.

thorwebdev (approved):
LGTM, thanks

david-plugge:
@cwierzbicki00 this should work:

```
const supabase = createMiddlewareSupabaseClient({ req, res }, {
    supabaseUrl: '...',
    supabaseKey: '...'
});
```

PR #444, opened 2023-02-09
[sveltekit] Updated typing to let server api routes call getServerSession

jennnx:

Fixed and updated typing.

The `getServerSession()` function in `auth-helpers` only accepts events of type `ServerLoadEvent`.
However, [api routes](https://kit.svelte.dev/docs/routing#server) in SvelteKit, which also only run on the server side, require this function as well to validate incoming requests.

But since api routes have an event type of [`RequestEvent`](https://kit.svelte.dev/docs/types#public-types-requestevent) (the parent type of `ServerLoadEvent`), the narrowly-defined auth helper will not allow this.

  
Screen Shot 2023-02-08 at 1 25 19 PM

## What is the new behavior?

Allow both server load events and server api routes to call the `getServerSession()` method by updating the typing.

## Additional context

I checked to make sure `RequestEvent` is only used in server-side functions. In addition, client-side request handlers expect the [NavigationEvent](https://kit.svelte.dev/docs/types#public-types-navigationevent) type, so i think we're good to go here.

### DIFF
--- a/.changeset/hip-moons-sip.md
+++ b/.changeset/hip-moons-sip.md
@@ -1,0 +1,5 @@
+---
+'@supabase/auth-helpers-sveltekit': patch
+---
+
+Updated typings to allow getServerSession to be called from server api routes
--- a/packages/nextjs/src/constants.ts
+++ b/packages/nextjs/src/constants.ts
@@ -1,2 +1,2 @@
 export const PKG_NAME = "@supabase/auth-helpers-nextjs";
-export const PKG_VERSION = "0.5.2";
+export const PKG_VERSION = "0.5.4";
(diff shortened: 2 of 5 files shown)


david-plugge:
The `getServerSession` function should only be used inside the root `+layout.server` file to make the session available in shared load functions. Instead `getSupabase(event)` should be used to get the supabaseClient and session (both are scoped to the request). This can also be found in the docs https://supabase.com/docs/guides/auth/auth-helpers/sveltekit#protecting-api-routes

This name is unfortunately poorly chosen and many users fall into this trap.

jennnx:
Ah, got it. That makes sense. Maybe in the future we could consider renaming it to like getServerLayoutSession or something then. Thanks for the help!

david-plugge:
We are working on a simpler api. You can have a look at it here https://github.com/supabase/auth-helpers/tree/sveltekit-overhaul/packages/sveltekit.

PR #449, opened 2023-02-13
Allow options.auth to be passed into createBrowser/ServerSupabaseClient

tonyxiao:

Feature

Currently we cannot pass any auth options when creating supabase client this way. For example `autoRefreshToken` on the client side

## What is the new behavior?

Allow user to specify default options for `auth` while overriding anything we need specifically in the helper. 

## Additional context

### DIFF
--- a/packages/shared/src/supabase-browser.ts
+++ b/packages/shared/src/supabase-browser.ts
@@ -1,6 +1,6 @@
-import { createClient, Session } from '@supabase/supabase-js';
+import { createClient, Session, SupabaseClientOptions } from '@supabase/supabase-js';
 import { parse, serialize } from 'cookie';
-import { CookieOptions, SupabaseClientOptionsWithoutAuth } from './types';
+import { CookieOptions } from './types';
 import { parseSupabaseCookie, stringifySupabaseSession } from './utils/cookies';
 import { isBrowser } from './utils/helpers';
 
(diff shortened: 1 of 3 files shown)


thorwebdev:
@tonyxiao supabase auth helpers sets the auth options that are needed for it to function, sorry that this isn't clear from the docs 😞 

You are not meant to change the auth options for the auth helpers client creation as that could mess with the intended behaviour.

tonyxiao:
@thorwebdev what about other options that should not impact the functionality, such as auth refresh access token?

thorwebdev:
`autoRefreshToken` does impact the functionality as it refreshes the access token on a timer which is not required with the auth helpers, as since supabase-js v2 the access token is checked (and refreshed if necessary) any time the library makes a request.

tonyxiao:
I see. But that introduces latency at request time which could be optimized if it was on a timer. Further is there any negative consequence of having `autoRefreshToken` on with the auth-helpers lib?

thorwebdev:
Sorry, `autoRefreshToken: false` is only set for the serverClient as this one isn't able to perform the refresh on timer: https://github.com/supabase/auth-helpers/blob/main/packages/shared/src/supabase-server.ts#L49

For the browserClient, auto refresh token mechanism is enabled (which is the default when `autoRefreshToken` is not specified) as it is required for realtime to work properly: https://github.com/supabase/auth-helpers/blob/main/packages/shared/src/supabase-browser.ts

So that's why we don't allow you to set the options, as it becomes quite difficult to narrow down problems if options aren't set as intended.

PR #455, opened 2023-02-22
Adds MFA factors into session

tonibardina:

Bug fix

When calling `stringifySupabaseSession` to build the session string, factors are not added to the list, this makes `getAuthenticatorAssuranceLevel` return the wrong `aal` data because this condition:

```typescript
//  inside getAuthenticatorAssuranceLevel definition
    const verifiedFactors =
      session.user.factors?.filter((factor: Factor) => factor.status === 'verified') ?? []
```

## What is the new behavior?

Factors are added to the session

### DIFF
--- a/.changeset/fresh-scissors-beg.md
+++ b/.changeset/fresh-scissors-beg.md
@@ -4,4 +4,4 @@
 
 Adds MFA factors into session
 
-Factors and identities were removed from session on [PR #350](https://github.com/supabase/auth-helpers/pull/350). When retrieving the `aal` data using ``
+Factors and identities were removed from session on [PR #350](https://github.com/supabase/auth-helpers/pull/350). When retrieving the `aal` data using `getAuthenticatorAssuranceLevel` wrong data is returned
--- a/.changeset/fresh-scissors-beg.md
+++ b/.changeset/fresh-scissors-beg.md
@@ -1,0 +1,7 @@
+---
+'@supabase/auth-helpers-shared': minor
+---
+
+Adds MFA factors into session
+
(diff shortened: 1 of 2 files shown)


tonibardina:
Steps to reproduce:
- Call `signInWithPassword` in an account with MFA enabled <- At this step the response from gotrue is saved as the session cookie (without factors)
- On auth state change make a redirect to another path (nextjs): `router.replace(/whatever)`
- Check `aal` with `getAuthenticatorAssuranceLevel`
- Since no factors are saved the response return `aal1` as the current and next level.

c.c @thorwebdev @david-plugge

david-plugge:
I wonder if it makes sense to rework the whole `stringify` method. I´m pretty sure the will be more things we have to add in the future so i think it´s time to add cookie splitting so we don´t have to worry about the actual session size.

tonibardina:
@david-plugge totally agree, but this might require adding changes all across the current flow. Since it might take some time I wonder if it might makes sense to add this quick fix to the current MFA flow till we implement a better solution? Also instead of saving all the factors we can at least save a flag to identify if there are any verified factors for that user.

thorwebdev:
Yes to both. Happy to get this in first and then look into reworking later. Thanks 💚

david-plugge (approved):
LGTM

J0:
Hey @david-plugge,

Hmm is there a approximate date which we have the sveltekit released planned for? Getting quite a few tickets from customers so hoping to get this merged soon.

Let us know!

david-plugge:
We are trying to get the release out by the end of tomorrow

PR #458, opened 2023-03-01
Feature: Add route handler package for app directory

dijonmusters:

feature

You need to create a Server Component Supabase Client in Route Handler functions

## What is the new behavior?

Export a new `createRouteHandlerSupabaseClient` function for use in Route Handler functions

### DIFF
--- a/.changeset/selfish-pears-applaud.md
+++ b/.changeset/selfish-pears-applaud.md
@@ -1,0 +1,5 @@
+---
+'@supabase/auth-helpers-nextjs': patch
+---
+
+Export function to create Supabase Client in Next.js Route Handler functions
--- a/packages/nextjs/src/index.ts
+++ b/packages/nextjs/src/index.ts
@@ -240,57 +240,5 @@ export function createServerComponentSupabaseClient<
   });
 }
 
-export function createRouteHandlerSupabaseClient<
-  Database = any,
-  SchemaName extends string & keyof Database = 'public' extends keyof Database
-    ? 'public'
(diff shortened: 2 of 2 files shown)


thorwebdev:
I'm bearish to introduce another client. Ideally we'd actually be boiling them back down into one server client. Can you elaborate why we need another one? 

Also, it looks like it's just the same as `createServerComponentSupabaseClient`, in that case couldn't we just do 

```js
export const createRouteHandlerSupabaseClient = createServerComponentSupabaseClient;
```

If it's just a nickname I mean. Or is there a functional difference?

dijonmusters:
> I'm bearish to introduce another client. Ideally we'd actually be boiling them back down into one server client. Can you elaborate why we need another one?
> 
> Also, it looks like it's just the same as `createServerComponentSupabaseClient`, in that case couldn't we just do
> 
> ```js
> export const createRouteHandlerSupabaseClient = createServerComponentSupabaseClient;
> ```
> 
> If it's just a nickname I mean. Or is there a functional difference?

No, that would work perfectly - just wanted it to be flexible enough to modify in case Next.js change the way Route Handlers work, but no need for preemptive flexibility 👍

thorwebdev (approved):
Still not a fan, but your call 😉

dijonmusters:
@thorwebdev and I have decided this is a good way to go while the app directory is still in beta. When they flick the switch to stable, we will do a major release of the auth helpers consolidating the app directory clients and renaming the pages-specific ones 👍

PR #460, opened 2023-03-02
[SvelteKit] Correctly type returned and cached client with generic param

nathancahill:

TS types bug fix with new SvelteKit client.

Returned client wasn't correctly narrowing `SchemaName` in `SupabaseClient`.

## What is the new behavior?

Returned type is correctly narrowed with generic parameters

## Additional context

The second change forces the cached client to return with the same type.

### DIFF
--- a/packages/sveltekit/src/supabaseLoadClient.ts
+++ b/packages/sveltekit/src/supabaseLoadClient.ts
@@ -84,10 +84,10 @@ export function createSupabaseLoadClient<
    * The event listener only runs in the browser.
    */
   onAuthStateChange?: (event: AuthChangeEvent, session: Session | null) => void;
-}) {
+}): SupabaseClient<Database, SchemaName> {
   const browser = isBrowser();
   if (browser && cachedBrowserClient) {
-    return cachedBrowserClient;
+    return cachedBrowserClient as SupabaseClient<Database, SchemaName>;
   }
 
   // this should never happen


david-plugge (approved):

david-plugge:
Thanks!

nathancahill:
Thanks for the awesome new Sveltekit implementation. I grabbed it directly since we were having a ton of minor bugs with the last one. The SvelteKit `fetch` usage is great.

Ran in to this other bug, you might want to be aware of since it affects SvelteKit more than other libraries because of the recommended `invalidate` in the top level +layout: https://github.com/supabase/supabase-js/issues/716

nathancahill:
Docs might be best updated to check the last state before invalidating:

```typescript
export const load: LayoutLoad = async (event) => {
	event.depends('supabase:auth');

	let lastAuthState: AuthChangeEvent | null = null;

	const supabase = createSupabaseLoadClient<Database, 'public'>({
		supabaseUrl: PUBLIC_SUPABASE_URL,
		supabaseKey: PUBLIC_SUPABASE_ANON_KEY,
		event: { fetch: event.fetch },
		serverSession: event.data.session,
		onAuthStateChange(event: AuthChangeEvent) {
			if (event !== lastAuthState) {
				lastAuthState = event;
				invalidate('supabase:auth');
			}
		}
	});
};
```

nathancahill:
Saved one more refresh by using the `event.session` like this:

```typescript
export const load: LayoutLoad = async (event) => {
	event.depends('supabase:auth');

	let lastAuthState: AuthChangeEvent | null = event.data.session ? 'SIGNED_IN' : null;

	const supabase = createSupabaseLoadClient<Database, 'public'>({
		supabaseUrl: PUBLIC_SUPABASE_URL,
		supabaseKey: PUBLIC_SUPABASE_ANON_KEY,
		event: { fetch: event.fetch },
		serverSession: event.data.session,
		onAuthStateChange(event: AuthChangeEvent) {
			if (event !== lastAuthState) {
				lastAuthState = event;
				invalidate('supabase:auth');
			}
		}
	});
};
```

PR #472, opened 2023-03-12
deprecate onAuthStateChange property

david-plugge:
closes https://github.com/supabase/auth-helpers/issues/470

### DIFF
--- a/packages/sveltekit/src/supabaseLoadClient.ts
+++ b/packages/sveltekit/src/supabaseLoadClient.ts
@@ -82,6 +82,9 @@ export function createSupabaseLoadClient<
   cookieOptions?: CookieOptions;
   /**
    * The event listener only runs in the browser.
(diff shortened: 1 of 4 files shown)


packages/sveltekit/src/supabaseLoadClient.ts
```diff
@@ -82,6 +82,9 @@ export function createSupabaseLoadClient<
   cookieOptions?: CookieOptions;
   /**
    * The event listener only runs in the browser.
+   * @deprecated DO NOT USE THIS
+   *
+   * use this instead: https://supabase.com/docs/guides/auth/auth-helpers/sveltekit#setting-up-the-event-listener-on-the-client-side
```
silentworks, line 87: We should probably inline the content from the link. So tell the user rather than redirecting them. We can add the link below that.

silentworks (approved):

PR #486, opened 2023-04-05
Remix example: Check fetcher state to avoid infinite loop

kellycampbell:
In some cases, an auth state change was causing an infinite loop of useEffect. This checks the state of the fetcher before calling submit to make sure it isn't already being submitted.

Bug fix in example

I didn't see this when running the simple example app, but ran into it with my own app which had no differences in the __supabase.tsx file. When clicking logout or completing login, the browser would freeze because the render was doing an infinite loop of the useEffect.

[Remix #4386](https://github.com/remix-run/remix/pull/4386)

## What is the new behavior?

Infinite loop doesn't occur.

### DIFF
--- a/examples/remix/app/routes/__supabase.tsx
+++ b/examples/remix/app/routes/__supabase.tsx
@@ -56,7 +56,7 @@ export const loader = async ({ request }: LoaderArgs) => {
 
 export default function Supabase() {
   const { env, session } = useLoaderData<typeof loader>();
-  const fetcher = useFetcher();
+  const refreshFetcher = useFetcher();
 
   // it is important to create a single instance of Supabase
   // to use across client components - outlet context 👇
@@ -70,10 +70,10 @@ export default function Supabase() {
     const {
       data: { subscription }
     } = supabase.auth.onAuthStateChange((event, session) => {
-      if (session?.access_token !== serverAccessToken) {
+      if (session?.access_token !== serverAccessToken && refreshFetcher.state === "idle") {
         // server and client are out of sync.
         // Remix recalls active loaders after actions complete
-        fetcher.submit(null, {
+        refreshFetcher.submit(null, {
           method: 'post',
           action: '/handle-supabase-auth'
         });
@@ -83,7 +83,7 @@ export default function Supabase() {
     return () => {
       subscription.unsubscribe();
     };
-  }, [serverAccessToken, supabase, fetcher]);
+  }, [serverAccessToken, supabase, refreshFetcher]);
 
   return (
     <>


joryphillips:
Note for internet searches/posterity: anyone using the useRevalidator() pattern and also hitting this infinite loop issue, also need to check for the idle state for it.

It might look like this:
```ts
  const revalidator = useRevalidator();
  const serverAccessToken = session?.accessToken;

  useEffect(() => {
    const {
      data: { subscription },
    } = supabase.auth.onAuthStateChange((_event, session) => {
      if (
        session?.access_token !== serverAccessToken &&
        revalidator.state === "idle"                            // <---- ADD THIS 
      ) {
        console.log("server and client are out of sync.");
        revalidator.revalidate();
      }
    });

    return () => {
      subscription.unsubscribe();
    };
  }, [serverAccessToken, supabase, revalidator]);
  ```

spy16:
I am running into this an infinite loop even after following this PR and this tutorial: https://www.youtube.com/watch?v=Viaed7XWCY8&ab_channel=Supabase

astonfuture:
> Note for internet searches/posterity: anyone using the useRevalidator() pattern and also hitting this infinite loop issue, also need to check for the idle state for it.
> 
> It might look like this:
> 
> ```ts
>   const revalidator = useRevalidator();
>   const serverAccessToken = session?.accessToken;
> 
>   useEffect(() => {
>     const {
>       data: { subscription },
>     } = supabase.auth.onAuthStateChange((_event, session) => {
>       if (
>         session?.access_token !== serverAccessToken &&
>         revalidator.state === "idle"                            // <---- ADD THIS 
>       ) {
>         console.log("server and client are out of sync.");
>         revalidator.revalidate();
>       }
>     });
> 
>     return () => {
>       subscription.unsubscribe();
>     };
>   }, [serverAccessToken, supabase, revalidator]);
> ```

I think every time you create the revalidator it generates it as a brand new object hence triggers the useEffect to run again. One suggestion was to use a useRef() to create the revalidator but I found it stopped the loop by just using destructuring  since we're defining a variable rather than creating a new object. 

const {revalidate} = useRevalidator()

joryphillips:
> I think every time you create the revalidator it generates it as a brand new object hence triggers the useEffect to run again. One suggestion was to use a useRef() to create the revalidator but I found it stopped the loop by just using destructuring since we're defining a variable rather than creating a new object.
> 
> `const {revalidate} = useRevalidator()`

Oh that is MUCH better, thank you for this!!

dijonmusters (approved):
LGTM! Thanks @david-plugge! 🙌

PR #498, opened 2023-04-11
feat: add pkce param to auth helpers

J0:

Introduces pkce param and updates supabase-js version in helpers


silentworks:
This is being handled on a different branch already.

PR #532, opened 2023-05-04
Fix Failed to parse cookie string in middleware

silentworks:

Bug fix

When using the `createMiddlewareSupabaseClient` inside a middleware the storage driver would throw an error due to the cookie being saved twice.

## What is the new behavior?

There is no longer an error and cookie is only saved once.

## Additional context

### DIFF
--- a/.changeset/brave-lizards-thank.md
+++ b/.changeset/brave-lizards-thank.md
@@ -1,0 +1,5 @@
+---
+'@supabase/auth-helpers-nextjs': patch
+---
+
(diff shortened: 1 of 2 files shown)


jlfwong:
Hi! We ran into this while working on https://streamlineclimate.com/. We have workarounds for now, but would love to have this fixed upstream. Is there an ETA on an npm publish that includes this change?

CC @douglasqian

silentworks:
This is already published on npm on the `.next` release. We are working to get a final release out soon, but currently we need to update the docs before we can make this a final release.

PR #543, opened 2023-05-16
WIP: Writable cookies

dijonmusters:
This is an experimental idea for how we can handle creating a Supabase Client in Route Handlers and Server Actions, now that they allow for writable cookies.


packages/nextjs/src/routeHandlerClient.ts
david-plugge, line 1: We should probably switch `createRouteHandlerSupabaseClient` and `createServerActionSupabaseClient` since the server actions api is not finalized and might change.

PR #547, opened 2023-05-17
Chore: Implement naming convention for createClient functions

dijonmusters:

Chore

1. createClient functions do not follow a clear convention
2. pages directory createClient functions are the default - no longer the recommended path

## What is the new behavior?

The App Router createClient functions follow a contextual convention as to where they are being used:

- createClientComponentClient
- createServerComponentClient
- createMiddlewareClient
- createRouteHandlerClient
- createServerActionClient

Pages directory createClient functions now contain `pages` in their name

- createPagesBrowserClient
- createPagesServerClient

---

Have also removed the `Supabase` part of each export, as this feels unnecessary.

## Additional context

We cannot consolidate all server-side functions into a single createServerClient function, as Server Components will not be able to `set` cookies, therefore, will always need to use middleware to refresh expired sessions: https://github.com/vercel/next.js/discussions/41745#discussioncomment-5198848

While `createRouteHandlerClient` and `createServerActionClient` are exactly the same function, I think it best to keep these as separate exports to simplify the DX when consuming them - if they were named the same, I think it would cause more confusion and require an unnecessarily deep understanding of the design decisions of the Next.js App Router.

This same justification goes for `createPagesBrowserClient` and `createClientComponentClient`. Having this clear separation between `pages` and `App Router` simplifies the DX.

### DIFF
--- a/packages/nextjs/src/serverComponentClient.ts
+++ b/packages/nextjs/src/serverComponentClient.ts
@@ -34,9 +34,7 @@ class NextServerComponentAuthStorageAdapter extends CookieAuthStorageAdapter {
 	}
 }
 
-export const createRouteHandlerSupabaseClient = createServerComponentSupabaseClient;
-
-export function createServerComponentSupabaseClient<
+export function createServerComponentClient<
 	Database = any,
 	SchemaName extends string & keyof Database = 'public' extends keyof Database
 		? 'public'
--- a/.changeset/six-eggs-search.md
+++ b/.changeset/six-eggs-search.md
@@ -1,0 +1,5 @@
+---
+'@supabase/auth-helpers-nextjs': patch
+---
+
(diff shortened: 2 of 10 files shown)


packages/nextjs/src/serverComponentClient.ts
```diff
@@ -34,9 +34,7 @@ class NextServerComponentAuthStorageAdapter extends CookieAuthStorageAdapter {
 	}
 }
 
-export const createRouteHandlerSupabaseClient = createServerComponentSupabaseClient;
```
dijonmusters, line 37: I think this one was accidentally left in from previous PR

silentworks: Yeah I think so.

silentworks (approved):

dijonmusters:
Awesome work! Thanks @silentworks! 🙌

PR #571, opened 2023-05-31
Chore: Extend config options to allow for multiple Supabase clients

dijonmusters:

Bug fix

Client Components could not create multiple Supabase clients.

## What is the new behavior?

A new config option for `isSingleton` is exposed on the `createClientComponentClient` function, to allow for multiple Supabase clients.

## Additional context

https://github.com/supabase/auth-helpers/issues/563

### DIFF
--- a/.changeset/unlucky-planes-think.md
+++ b/.changeset/unlucky-planes-think.md
@@ -1,0 +1,5 @@
+---
+'@supabase/auth-helpers-nextjs': patch
+---
+
+Allow user to opt out of singleton pattern for Client Components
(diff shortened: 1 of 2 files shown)


.changeset/unlucky-planes-think.md
J0, line 1: Is the name of the change set supposed to have any significance? Not sure how the .changeset files are used

J0 (approved):

.changeset/unlucky-planes-think.md
dijonmusters: No, they are autogenerated for the purposes of releases. @silentworks can probably explain this in more depth 👍

PR #591, opened 2023-06-21
Handle initialSession changes in context provider

### LINKED ISSUE #586: `useUser` is always `null` on first load
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

We’re using `useUser` to conditionally display user data. The `initialSession` is passed via server side props to `SessionContextProvider` in the `_app.jsx`, however, the context isn’t correctly passing this value down to the hook.

## To Reproduce

Below is an example of the protected route users are redirected to after logging in. The `initialSession` prop is passed through as expected, however, the `user` value from the hook is `null`.

```jsx
import { createPagesServerClient } from "@supabase/auth-helpers-nextjs";
import { useUser } from "@supabase/auth-helpers-react";

export default function Protected({ initialSession }) {
  const user = useUser();

  return JSON.stringify({
    initialSessionUser: initialSession?.user,
    hookUser: user, // null
  });
}

export async function getServerSideProps(context) {
  const supabase = createPagesServerClient(context);

  const {
    data: { session },
  } = await supabase.auth.getSession();

  if (!session) {
    return {
      redirect: {
        destination: "/login",
        permanent: false,
      },
    };
  }

  return {
    props: {
      initialSession: session,
    },
  };
}
```

## Expected behavior

That the `useUser` hook would return `initialSession.user` from the context.

## System information

- OS: macOS
- Version of supabase-js: 2.4.0
- Version of Node.js: 18

joshuabaker:
Resolves #586

This is a bug/logic fix for how `initialSession` is handled in the `SessionContextProvider`.

At present, `initialSession` is passed into `useState`, which means that it is discarded after the first render. The issue here is that this is often on a login page where it will likely be `undefined`, meaning any session subsequently created on the server side is discarded.

## What is the new behavior?

I’ve just added a `useEffect` that sets the session state if `initialSession` is available and `session` is not.

### DIFF
--- a/.changeset/flat-seas-attack.md
+++ b/.changeset/flat-seas-attack.md
@@ -1,0 +1,5 @@
+---
+'@supabase/auth-helpers-react': patch
+---
+
+fix: Handle initialSession changes in context provider
--- a/packages/react/src/components/SessionContext.tsx
+++ b/packages/react/src/components/SessionContext.tsx
@@ -56,6 +56,12 @@ export const SessionContextProvider = ({
 	const [error, setError] = useState<AuthError>();
 
 	useEffect(() => {
+		if (!session && initialSession) {
+			setSession(initialSession)
+		}
+	}, [session, initialSession]);
+	
+	useEffect(() => {
 		let mounted = true;
 
 		async function getSession() {


silentworks (approved):
Sorry it too me so long to get around reviewing this. Thank you for this PR.

PR #594, opened 2023-06-24
feat: add solidstart helpers package

caseybaggz:

Adds new `solidstart` package and example.

There is none! 🙈 It's time for the solid community to get what they deserve... 👊 a helper for the Solid Start apps! 😄 

Please link any relevant issues here.

n/a

## What is the new behavior?

- Creates new solidstart workspace
- Adds new solidstart example using workspace build

## Additional context

Right now Supabase is getting a ton of attention in the SolidJS Discord server #solid-start channel around server rendering, auth, etc. This package will help us have the tools we need to successfully show how to connect SB using Auth and server rendering within a Solid Start app. I purposefully left some `console.log` in the example app which was very helpful to me in figuring out this solution, so I thought it might be useful to others as well.

It also updates the helpers package to use the new `exports` field because SS apps use Vite which only supports ESM modules.


silentworks:
It's best to create a separate repository/library for this similar to how the Nuxt team has done there as non of the maintainers of this project uses Solid or knows it well enough to maintain this going forward.

PR #600, opened 2023-07-04
feat: Add export fields to package.json

lachlancollins:

- Expands bundler/TS support, corrects https://publint.dev/ warnings

For @supabase/auth-helpers-sveltekit:
```This entrypoint has types at ./dist/index.d.ts but it is not exported from pkg.exports. Consider adding it to pkg.exports["."].types to be compatible with TypeScript's "moduleResolution": "bundler" compiler option.```

## What is the new behavior?

- Warnings removed :)
- The other packages didn't have warnings but I figured I should add `exports` while I'm at it

### DIFF
--- a/packages/sveltekit/README.md
+++ b/packages/sveltekit/README.md
@@ -56,11 +56,6 @@ export const handle: Handle = async ({ event, resolve }) => {
   };
 
   return resolve(event, {
-    /**
-     * There´s an issue with `filterSerializedResponseHeaders` not working when using `sequence`
-     *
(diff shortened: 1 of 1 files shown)


silentworks:
Please make this PR only for the sveltekit package and not the others as this might lead to side effects for the other packages. I'll get it merged in once this change has been made.

lachlancollins:
@silentworks no worries! I accidentally closed this PR trying to merge changes from main - the new PR is #608

PR #609, opened 2023-07-16
feat: allow promises in CookieAuthStorageAdapter

csenio:

feature / chore

`CookieAuthStorageAdapter` doesn't await the response of `this.getCookie(key)`.
This limits how much it can be extended as methods like `chrome.cookies.getAll` always run asynchronously.

## What is the new behavior?

`getItem` in the `CookieAuthStorageAdapter` is now an async function that `awaits` the result of `this.getCookie(key)`. 
This should not cause any changes in behaviour and is in line with how `getItem` works in the goTrueClient storage adapter.

## Additional context

My personal use case is that I'm making an adapter to use supabase auth in my extension service worker and I prefer extending `CookieAuthStorageAdapter` over `StorageAdapter`. I figured other people trying to extend it will run into the same issue eventually so making a PR.

### DIFF
--- a/packages/shared/src/cookieAuthStorageAdapter.ts
+++ b/packages/shared/src/cookieAuthStorageAdapter.ts
@@ -14,12 +14,14 @@ export abstract class CookieAuthStorageAdapter implements StorageAdapter {
 		};
 	}
 
-	protected abstract getCookie(name: string): string | undefined | null;
-	protected abstract setCookie(name: string, value: string): void;
+	protected abstract getCookie(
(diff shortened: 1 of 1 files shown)


mandarini:
Thank you for your PR and for contributing to Supabase @csenio ! I will be closing this PR since we have moved the functionality of `auth-helpers` into the [@supabase/ssr](https://github.com/supabase/ssr) package. Please direct your contributions there. Thank you for understanding and for being a member of our community!

PR #630, opened 2023-08-30
Feature: Add `ssr` package to configure cookies for browser or server clients

dijonmusters:

feature

Each framework implements their own very specific version of the same primitive functions - configure a Supabase client to use cookies for browser and server.

Furthermore, the browser client is reimplemented in the same way across each framework - the browser's `document` does not change based on framework!

## What is the new behavior?

These functions are exposed from the new `ssr` package, requiring just the bits that need to be implemented for each framework. The cookie-based browser client should be usable as is across any framework - the browser's `document` does not change based on framework!. As for the server, we could get rid of framework-specific libraries entirely, and just have documentation for how to `get`, `set` and `delete` cookies for each framework.

**Next.js**
```
import { createServerClient } from '@supabase/ssr'

const supabase = createServerClient(
  supabaseUrl,
  supabaseKey,
  {
    cookies: {
      get: (key) => cookies.get(key),
      set: (key, value, options) => cookies.set(key, value, options),
      delete: (key) => cookies.set(key, '', { maxAge: -1 })
    }
  }
)
```

**Astro**
```
import { createServerClient } from '@supabase/ssr'

const supabase = createServerClient(
  supabaseUrl,
  supabaseKey,
  {
    cookies: {
      get: (key) => Astro.cookies.get(key),
      set: (key, value, options) => Astro.cookies.set(key, value, options),
      delete: (key) => Astro.cookies.remove(key)
    }
  }
)
```

**Oak**
```
import { createServerClient } from '@supabase/ssr'

const supabase = createServerClient(
  supabaseUrl,
  supabaseKey,
  {
    cookies: {
      get: (key) => context.request.cookies.get(key),
      set: (key, value, options) => context.response.cookies.set(key, value, options),
      delete: (key) => context.response.cookies.set(key, '', { maxAge: -1 })
    }
  }
)
```

These improved primitives could potentially be integrated back into `supabase-js`, since they are really just a bit of configuration for auth storage. This, coupled with rewritten docs, could greatly simplify using Supabase across any web framework!

Ideally, the `supabase-js` implementation would look something like:

**localStorage**
```
import { createClient } from '@supabase/supabase-js'

const supabase = createClient(
  supabaseUrl,
  supabaseKey
)
```

**cookie-based storage**
```
import { createClient } from '@supabase/supabase-js'

const supabase = createClient(
  supabaseUrl,
  supabaseKey,
  {
    cookies: {
      get: (key) => cookies.get(key),
      set: (key, value, options) => cookies.set(key, value, options),
      delete: (key) => cookies.set(key, '', { maxAge: -1 })
    }
  }
)
```

**custom supabase client**
```
import { createClient } from '@supabase/supabase-js'

const supabase = createClient(
  supabaseUrl,
  supabaseKey,
  {
    auth: {
      storage: {
        ...
      }
    }
  }
)
```


packages/ssr/src/utils/cookies.ts
```diff
@@ -0,0 +1,77 @@
+import { Session } from '@supabase/supabase-js';
+import { parse, serialize } from 'cookie';
+import { base64url } from 'jose';
+
+export { parse as parseCookies, serialize as serializeCookie };
+
+/**
+ * Based on the environment and the request we know if a secure cookie can be set.
+ */
+export function isSecureEnvironment(headerHost?: string | string[]) {
```
dijonmusters, line 10: @silentworks Is this function still needed? Doesn't look like it is called anywhere

packages/ssr/src/index.ts
```diff
@@ -0,0 +1,12 @@
+export * from './createBrowserClient';
+export * from './createServerClient';
+export * from './types';
+
+export {
```
dijonmusters, line 5: @silentworks It looks like all we are excluding here is the `isSecureEnvironment` function (see other comment). If we removed that function, we could just `export * from './utils';`

packages/ssr/src/createBrowserClient.ts
```diff
@@ -0,0 +1,114 @@
...
+							...cookieOptions
+						});
+					}
+				}
+			}
+		}
+	};
+
+	// Overwrites default client config with any user defined options
+	const clientOptions = mergeDeepRight(
```
dijonmusters, line 81: @silentworks this `mergeDeepRight` utility makes the default configuration above much more readable. What are your thoughts?

packages/ssr/CHANGELOG.md
```diff
@@ -0,0 +1 @@
+# ssr
```
dijonmusters, line 1: @silentworks not sure if this file is created if it does not exist, when we do the changeset stuff

packages/ssr/src/utils/cookies.ts
dijonmusters: Removed 👍

packages/ssr/src/index.ts
dijonmusters: Have removed function and exported all utils

packages/ssr/package.json
```diff
@@ -0,0 +1,48 @@
...
+		"Svelte"
+	],
+	"author": "Supabase",
+	"license": "MIT",
+	"bugs": {
+		"url": "https://github.com/supabase/auth-helpers/issues"
+	},
+	"homepage": "https://github.com/supabase/auth-helpers#readme",
+	"devDependencies": {
+		"@supabase/supabase-js": "2.19.0",
```
dijonmusters, line 34: Does this need to be a `devDependency` and a `peerDependency`?

packages/ssr/src/createBrowserClient.ts
```diff
@@ -60,13 +66,15 @@ export function createBrowserClient<
 					if (isBrowser()) {
 						document.cookie = serialize(key, value, {
 							httpOnly: false,
```
dijonmusters, line 68: We could also move `httpOnly: false` to our DEFAULT_COOKIE_OPTIONS`

packages/ssr/CHANGELOG.md
silentworks: Yeah this gets created automatically by the changeset stuff.

packages/ssr/src/createBrowserClient.ts
silentworks: Good idea.

packages/ssr/package.json
```diff
@@ -0,0 +1,45 @@
...
+	"bugs": {
+		"url": "https://github.com/supabase/auth-helpers/issues"
+	},
+	"homepage": "https://github.com/supabase/auth-helpers#readme",
+	"devDependencies": {
+		"@supabase/supabase-js": "2.33.1",
+		"@types/cookie": "^0.5.1",
+		"@types/ramda": "^0.29.3",
+		"cookie": "^0.5.0",
+		"ramda": "^0.29.0",
```
kiwicopple, line 38: these ones seem like full depenencies (not devDependencies)? Perhaps not, just flagging in case

silentworks: As full dependencies they will get installed as individual packages, as dev dependencies the transpiler will treeshake them and include only the code from these libraries that are being used by this package.

silentworks: Although this makes a huge difference in bundle size. Will move these to `dependencies` instead of `devDependencies`.

silentworks (approved):

PR #632, opened 2023-09-06
Wrap createServerComponentClient in React.cache

tom-sherman:
This ensures you only create the supabase client once per request.

### DIFF
--- a/packages/nextjs/package.json
+++ b/packages/nextjs/package.json
@@ -29,6 +29,7 @@
 	"homepage": "https://github.com/supabase/auth-helpers/tree/main/packages/nextjs#readme",
 	"devDependencies": {
 		"@supabase/supabase-js": "2.19.0",
+		"@types/react": "^18.2.21",
 		"@types/set-cookie-parser": "^2.4.3",
 		"config": "workspace:*",
 		"next": "^13.4.19",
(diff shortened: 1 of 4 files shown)


tom-sherman:
There is a build error right now, the dts build fails because it doesn't know how to find @types/react/experimental. I'm not sure why this would be failing considering I've added it to dev dependencies. Does it need to be a peer?

remorses:
Using cache on this function has no effect because the argument is an object, it will be different every call

tom-sherman:
Damn, you're right.

I think this can be fixed by moving cookies into it's own argument and documenting that the options should be defined outside of render. Not sure if it's worth the breaking change.

Maybe we can add a curried form as a non breaking change? (function name to be bikeshedded)

```ts
const createServerComponentClientFactory = (options: Options) =>
  cache((cookies: Cookies) =>
    createServerComponentClient({ cookies }, options),
  );
```

Of course this function would need to be initialised once, but I think this is common with these kinds of clients.

Another options would be to stop accepting cookies as a param and just call `cookies()` inside, I'm not sure why this is wasn't done in the first place tho - maybe there's a constraint I'm not seeing.

remorses:
In my projects I wrap functions so the argument gets serialized before using cache using this package https://www.npmjs.com/package/fast-stable-stringify

But this function is cheap enough that cache shouldn’t be needed at all imo

tom-sherman:
It's not the work this function does that's the issue, it's that creating multiple supabase clients uses more memory and can potentially lead to duplicate fetches where they're not expected.

PR #651, opened 2023-10-12
fix: include cookieOptions for deleteCookie

dogukanakkaya:

Adds `this.cookieOptions` when calling `deleteCookie`

If cookie is set by a specific domain it cannot be deleted since `deleteCookie` only specifiy `maxAge` as option and not others like `domain`.


dijonmusters:
Great catch! Thanks @dogukanakkaya! 🙌

PR #653, opened 2023-10-20
Add cookie chunker code

silentworks:

Add string chunking for larger JWT to save into multiple cookies

If the JWT is too large, you can't use the auth-helpers.

## What is the new behavior?

Larger JWT gets saved into multiple cookies.

## Additional context


examples/sveltekit/src/routes/(app)/+page.server.ts
```diff
@@ -99,7 +99,7 @@ export const actions = {
 			}
 		});
 
-		if (!error) {
```
kamilogorek, line 102: 🥲

examples/sveltekit/src/routes/(app)/+page.svelte
```diff
@@ -27,7 +27,7 @@
 		{/if}
 
 		{#if $page.url.searchParams.get('auth-type') === 'magiclink'}
-			<form action="/?/create-magiclink" method="post" use:enhance={handleSubmit}>
```
kamilogorek, line 30: 🥲 x2

packages/shared/src/chunker.ts
```diff
@@ -0,0 +1,49 @@
...
+export function createChunks(key: string, value: string) {
+	// check the length and then split
+	const chunkCount = Math.ceil(value.length / MAX_CHUNK_SIZE);
+
+	if (chunkCount === 1) {
+		return value;
+	}
+
+	const chunks: Chunk[] = [];
+	const values = value.match(new RegExp('.{1,' + MAX_CHUNK_SIZE + '}', 'g'));
```
hf, line 19: Move the regexp to a constant, as creating a new regexp every time is very slow.

packages/shared/src/chunker.ts
```diff
@@ -0,0 +1,49 @@
...
+
+	values?.forEach((value, i) => {
+		const name: string = `${key}.${i}`;
+		chunks.push({ name, value });
+	});
```
kamilogorek, line 24: You can also write this without regexp, but I don't mind which one you use.
```suggestion
	let cookie = Array.from(value);
  while (cookie.length) {
    chunks.push({
      name: `${key}.${chunks.length}`,
      value: cookie.splice(0, MAX_CHUNK_SIZE)
    });
  }
```

packages/shared/src/chunker.ts
```diff
@@ -0,0 +1,49 @@
+// No clue why, but 3600 matches 4kb in the browser
```
hf, line 1: These tabs are huge no? :D

hf: Just in general though I'm not super clear how chunking works? Could you document the approach in the file itself?

Also I would prefer if this went into gotrue-js, I can see it being generally useful.

packages/shared/src/chunker.ts
```diff
@@ -0,0 +1,49 @@
...
+	value: string;
+}
+
+// Get cookie as chunks to set in the browser
+export function createChunks(key: string, value: string) {
+	// check the length and then split
+	const chunkCount = Math.ceil(value.length / MAX_CHUNK_SIZE);
+
+	if (chunkCount === 1) {
+		return value;
```
kamilogorek, line 15: I'd either skip it entirely and let it flow through the rest of code, or return `[value]` instead, this way we dont need to branch out to `Array.isArray` later on.

packages/shared/src/cookieAuthStorageAdapter.ts
```diff
@@ -42,10 +49,38 @@ export abstract class CookieAuthStorageAdapter implements StorageAdapter {
 		let session: Session = JSON.parse(value);
 		const sessionStr = stringifySupabaseSession(session);
 
-		this.setCookie(key, sessionStr);
+		// split session string before setting cookie
+		const sessionChunks = createChunks(key, sessionStr);
```
kamilogorek, line 53: See comment about `chunks.length === 1` above.

examples/sveltekit/src/routes/(app)/+page.server.ts
silentworks: Sorry that this unrelated change got included.

examples/sveltekit/src/routes/(app)/+page.svelte
silentworks: Another unrelated change, this should have been in a different PR but it was a necessary change during testing.

examples/sveltekit/src/routes/(app)/+page.server.ts
kamilogorek: No no, it's not about change being unrelated. It's the fact that it wasn't caught before and happily lived there :D

packages/shared/src/chunker.ts
silentworks: Yeah it would make sense if this was in `gotrue-js`. Would it be ok to get this released in the auth-helpers first and then work our way back to `gotrue-js` after?

examples/sveltekit/src/routes/(app)/+page.server.ts
silentworks: Yeah it probably wouldn't have been caught since the examples here are only used by those who work on this library and not everyone uses SvelteKit. It's generally the Next.js version that gets checked the most.

packages/shared/src/chunker.ts
silentworks: Wouldn't this be a bit slower since it would convert all characters of the value into a string?

kamilogorek: It'd definitely not be as slow as the regexp, but it's just a matter of taste anyway IMO. It's not a very hot code path.

packages/shared/src/chunker.ts
```diff
@@ -0,0 +1,53 @@
...
+	}
+) {
+	let values: string[] = [];
+	for (let i = 0; ; i++) {
+		const chunkName = `${key}.${i}`;
+		const chunk = retrieveChunk(chunkName);
+
+		if (!chunk) {
+			break;
+		}
```
hf, line 47: Ah okay this is a smart way to reconstruct it.

packages/shared/src/chunker.ts
```diff
@@ -0,0 +1,53 @@
...
+// Get fully constructed chunks
+export function combineChunk(
+	key: string,
+	retrieveChunk: (name: string) => string | null | undefined = () => {
+		return null;
+	}
+) {
+	let values: string[] = [];
+	for (let i = 0; ; i++) {
+		const chunkName = `${key}.${i}`;
```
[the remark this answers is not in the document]
silentworks: This is a good candidate for unit tests. Let me add some tests to the repo for these functions.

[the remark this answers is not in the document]
silentworks: I can confirm they are reassembled in the correct order. I've added some unit tests to make sure this is also correct.

packages/shared/src/chunker.ts
```diff
@@ -1,16 +1,25 @@
 // No clue why, but 3600 matches 4kb in the browser
-const MAX_CHUNK_SIZE = 3600;
-const re = new RegExp('.{1,' + MAX_CHUNK_SIZE + '}', 'g');
+let MAX_CHUNK_SIZE = 3600;
```
kamilogorek, line 2: I'd rewrite this whole thing to:

```js
const MAX_CHUNK_SIZE = 3600;
const MAX_CHUNK_REGEXP = createChunkRegExp(MAX_CHUNK_SIZE);

function createChunkRegExp(chunkSize: number) {
  return new RegExp('.{1,' + MAX_CHUNK_SIZE + '}', 'g');
}

export function createChunks(key: string, value: string, chunkSize?: number): Chunk[] {
  const re = chunkSize !== undefined ? createChunkRegExp(chunkSize) : MAX_CHUNK_REGEXP;
  
  // ...
}
```

It preserves a constant default, no type checking, and allows for optional overrides for tests.
It also doesn't export unnecessary, test-only functions.

kamilogorek (approved):

PR #654, opened 2023-10-23
Chore: make cookie functions optional in server client

dijonmusters:

Chore

Cookie functions are mandatory when creating s Server client

## What is the new behavior?

Cookie functions are optional when creating s Server client

## Additional context

Next.js Server Components cannot set or remove cookies. It doesn't make sense to declare these empty functions to be able to use a Supabase client in a Server Component

### DIFF
--- a/packages/ssr/src/types.ts
+++ b/packages/ssr/src/types.ts
@@ -2,13 +2,8 @@ import type { CookieSerializeOptions } from 'cookie';
 
 export type CookieOptions = Partial<CookieSerializeOptions>;
 export type CookieOptionsWithName = { name?: string } & CookieOptions;
-export type ServerCookieMethods = {
+export type CookieMethods = {
 	get?: (key: string) => Promise<string | null | undefined> | string | null | undefined;
-	set?: (key: string, value: string, options?: CookieOptions) => Promise<void> | void;
-	remove?: (key: string, options?: CookieOptions) => Promise<void> | void;
-};
-export type BrowserCookieMethods = {
-	get?: (key: string) => Promise<string | null | undefined> | string | null | undefined;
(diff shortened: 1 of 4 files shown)


silentworks (approved):

packages/ssr/src/types.ts
```diff
@@ -3,9 +3,9 @@ import type { CookieSerializeOptions } from 'cookie';
 export type CookieOptions = Partial<CookieSerializeOptions>;
 export type CookieOptionsWithName = { name?: string } & CookieOptions;
 export type ServerCookieMethods = {
```
silentworks, line 5: I wonder if now we should just have on set of types since the `ServerCookieMethods` and `BrowserCookieMethods` are exactly the same.

dijonmusters: Good call! Have consolidated them into one 👍

packages/ssr/src/types.ts
```diff
@@ -2,13 +2,8 @@ import type { CookieSerializeOptions } from 'cookie';
 
 export type CookieOptions = Partial<CookieSerializeOptions>;
 export type CookieOptionsWithName = { name?: string } & CookieOptions;
-export type ServerCookieMethods = {
-	get: (key: string) => Promise<string | null | undefined> | string | null | undefined;
-	set: (key: string, value: string, options?: CookieOptions) => Promise<void> | void;
-	remove: (key: string, options?: CookieOptions) => Promise<void> | void;
-};
-export type BrowserCookieMethods = {
+export type CookieMethods = {
```
dijonmusters, line 5: Also realised the `options` parameter should not be optional. Both `createServerClient` and `createBrowserClient` pass in the default cookie options when they call these functions

PR #669, opened 2023-11-03
Fix: add cookie chunking as default to ssr package

dijonmusters:

Fix

Some OAuth signin methods are failing due to size of cookie being too big

## What is the new behavior?

Implements cookie chunking to ensure size of each cookie is not too big

### DIFF
--- a/packages/ssr/src/utils/chunker.ts
+++ b/packages/ssr/src/utils/chunker.ts
@@ -7,7 +7,7 @@ function createChunkRegExp(chunkSize: number) {
 	return new RegExp('.{1,' + chunkSize + '}', 'g');
 }
 
-const MAX_CHUNK_SIZE = 1600;
+const MAX_CHUNK_SIZE = 3600;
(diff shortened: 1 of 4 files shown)


packages/ssr/src/utils/chunker.ts
```diff
@@ -7,15 +7,14 @@ function createChunkRegExp(chunkSize: number) {
 	return new RegExp('.{1,' + chunkSize + '}', 'g');
 }
 
-const MAX_CHUNK_SIZE = 3600;
+const MAX_CHUNK_SIZE = 1600;
```
silentworks, line 10: We need to remember to set this back to 3600 as this was only for testing.

dijonmusters: I was wondering about that!

silentworks (approved):

mb21:
Am I right in assuming that we need the chunking only because we store so much stuff in the cookies, much of it redundant information? Wouldn't it be enough to store only the access token and refresh token? All the other info can be derived from that server-side?

PR #677, opened 2023-11-04
Revert cookie name to storage key change

silentworks:


binury:
This seems important to have 😕

jorgebaralt:
why is this gone?

PR #680, opened 2023-11-09
Fixed chunker chunk length calculation

dvvolynkin:

Bug fix

[Cookie is too large](https://github.com/supabase/auth-helpers/issues/643)
Value not encoded before splitting. 
Sometimes, this causes too big chunks because the encoded cookie value length is longer than not encoded.

## What is the new behavior?
Added splitting by encoded value


silentworks:
Thank you for this PR. We just need to test it out and make sure the unit tests for the chunker is still passing and then we can get this merged in.

silentworks:
Unit tests are failing on this change, the chunk length returned is always 1 which cannot be correct.

packages/ssr/src/utils/chunker.ts
```diff
@@ -4,7 +4,7 @@ interface Chunk {
 }
 
 function createChunkRegExp(chunkSize: number) {
-	return new RegExp('.{1,' + chunkSize + '}', 'g');
+	return new RegExp(`(.{1,${chunkSize}})(?=%[0-9A-Fa-f]{2}|$)`, 'g');
```
silentworks, line 7: Why this change to the Regex?

dvvolynkin: Yep, it was a mistake
Changing it back

dvvolynkin:
@silentworks 
Changed regexp back, now tests are passing 

Sorry that haven't checked the tests in the first commit.

dvvolynkin:
@silentworks 
Added test for encoded string and fixed bug in my code

silentworks:
@dvvolynkin you've made too many logic change in this code, please revert the logic to what it was before as your changes are complicating the code.

silentworks (changes requested):
You've made too many logic change in this code, please revert the logic to what it was before as your changes are complicating the code.

dvvolynkin:
@silentworks okay

I revert all logic of chunker and leave a test that is not passing.
You can implement a logic in your style.

dvvolynkin:
1. Cookie value should be splitted into chunks that length can fit 4096 symbols **after encoding**. 
2. If you use the initial regexp on the encoded string then you will face that your chunks can be broken because of % sequence.

kangmingtay:
hi @dvvolynkin, thanks for working with @silentworks to hash out the possible solutions we can take to chunk the cookies correctly. just an update, we've decided to go with this approach: https://github.com/supabase/auth-helpers/pull/726 so i'll be closing this PR

PR #730, opened 2024-01-26
fix: use cookie name as storage key

kangmingtay:
* Currently, setting a custom cookie name via the `cookieOptions` breaks the client because the cookies will be set with the custom name but `getItem` uses the `storageKey` value to retrieve the cookie


J0 (approved):

hf (approved):

binury:
@kangmingtay why is it even possible to set differing values for storageKey & cookie.name if it is imperative that they are identical?

PR #736, opened 2024-01-31
chore: add missing changesets

kangmingtay:
* Changes made to cookie chunking in ssr package: #726 
* Changes made in server-side storage adapters: #722


J0 (approved):
lgtm, thanks for putting this together

/nit we should consider renaming the changeset in future. Should be fine for now if we are going to consume via `changeset version` soon

Issue #717, opened 2024-01-05
cookieOption field "name" is not correctly working


cp-20:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [ ] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

I used `createPagesBrowserClient` for my application and set cookieOption to the below but stored cookie didn't change its name.

```
{
	name: 'some-cookie-name',
	secure: true,
	sameSite: "Lax",
	domain: "",
	path: "/",
}
```

I suspect that https://github.com/supabase/auth-helpers/commit/f7e5c2dc01248868e763b93a12c0539eb4bcc615 makes this bug. Base cookie name passed to `createClient` of `@supabase/supabase-js` comes from `option.auth.storageKey` before this commit, but `storageKey` field is removed and nothing other is changed after it. Please consider reverting it or take another solution.

fnimick:
This is blocking a migration from @supabase/auth-helpers-sveltekit (where the name option works) to ssr (where the name option no longer works). I can confirm that the name option has no effect on the cookie used for auth data.

kangmingtay:
@cp-20 @fnimick we fixed this in the supabase/ssr package in #730 - i'll take a look at the other packages to see if the same bug is present there too

binury:
> @cp-20 @fnimick we fixed this in the supabase/ssr package in #730 - i'll take a look at the other packages to see if the same bug is present there too

I could be mistaken but #730 change applies to the serverClient but not browserClients which will still use `sb-${new URL(this.authUrl).hostname.split('.')[0]}-auth-token`. Without that, it's only going to be making "anon" requests.

_Apparently_ this did exist at one point https://github.com/supabase/auth-helpers/pull/677 but was reverted for unclear reason(s).

jorgebaralt:
Any reason it got reverted? I really need this

binury:
> Any reason it got reverted? I really need this 

as a workaround, you need to manually set the storageKey and the cookie.name to the same value

jorgebaralt:
> > Any reason it got reverted? I really need this
> 
> as a workaround, you need to manually set the storageKey and the cookie.name to the same value. or, don't use the beta ssr auth-helper

Thanks for the help. 

[...]

PR #746, opened 2024-03-05
fix: allow setting custom cookie name in createBrowserClient

kangmingtay:
* Address [the comment](https://github.com/supabase/auth-helpers/issues/717#issuecomment-1962251408) raised in #717 
* `createBrowserClient` in the ssr package also doesn't seem to respect the custom cookie name used

### DIFF
--- a/.changeset/proud-coins-help.md
+++ b/.changeset/proud-coins-help.md
@@ -1,0 +1,5 @@
+---
+'@supabase/ssr': minor
+---
+
+fix custom cookie options in browser client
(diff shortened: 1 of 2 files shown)


J0 (approved):

jorgebaralt:
Any reason this is not merged yet? @J0 @kangmingtay

J0:
Hey @jorgebaralt,

Thanks for checking in. It's been a little busy - we'll need some time to test this to confirm it works as expected. If anyone has the bandwidth to check that this works as expected it might helps speed the review process. We will still need to test it ourselves as well though

PR #751, opened 2024-03-14
fix: bump supabase-js to match lastest types.

NicolasMontone:

Fix. right now types definitions are not working when using ssr library added https://github.com/supabase/storage-js/pull/190
image
When in the console when we log the data is -> 
image

I'm not having type completation


J0 (approved):

J0:
Thank you

PR #760, opened 2024-04-05
fix: small improvements/fixes to cookie chunker in ssr

hf:
Fixes/improves some edge cases in the cookie chunker for `ssr`:

1. `deleteChunks` now attempts to delete both the key and potential leftover chunks, instead of deleting the first thing it notices. This should decrease the total size of cookies by removing orphaned chunks, and also is a bit safer in situations where leftover chunks are mistaken as the continuation of other chunks.
2. Avoid using parallel cookie setting in `setItem`, and instead do it serially.
3. `setItem` first clears all chunks before setting the new ones.

### DIFF
--- a/packages/ssr/src/createServerClient.ts
+++ b/packages/ssr/src/createServerClient.ts
@@ -46,6 +46,26 @@ export function createServerClient<
 		};
 	}
 
+	const deleteAllChunks = async (key: string) => {
+		await deleteChunks(
+			key,
+			async (chunkName) => {
+				if (typeof cookies.get === 'function') {
+					return await cookies.get(chunkName);
+				}
+			},
+			async (chunkName) => {
+				if (typeof cookies.remove === 'function') {
+					return await cookies.remove(chunkName, {
+						...DEFAULT_COOKIE_OPTIONS,
+						...cookieOptions,
+						maxAge: 0
+					});
+				}
+			}
+		);
+	};
+
 	const cookieClientOptions = {
 		global: {
 			headers: {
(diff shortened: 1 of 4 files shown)


packages/ssr/src/createServerClient.ts
```diff
@@ -70,17 +70,14 @@ export function createServerClient<
...
-									...DEFAULT_COOKIE_OPTIONS,
-									...cookieOptions,
-									maxAge: DEFAULT_COOKIE_OPTIONS.maxAge
-								});
-							}
-						})
-					);
+
+					for (let i = 0; i < chunks.length; i += 1) {
+						await cookies.set(chunk.name, chunk.value, {
```
hf, line 75: Needs typeof check

hf:
One confirmed support ticket that is demonstrating the issues resolved by these changes.

J0 (approved):
Seems sane at brief glance, approving for now to unblock and will test in the morrow if not yet merged

packages/ssr/src/createServerClient.ts
```diff
@@ -69,18 +89,22 @@ export function createServerClient<
...
-							if (typeof cookies.set === 'function') {
-								await cookies.set(chunk.name, chunk.value, {
-									...DEFAULT_COOKIE_OPTIONS,
-									...cookieOptions,
-									maxAge: DEFAULT_COOKIE_OPTIONS.maxAge
-								});
-							}
-						})
-					);
+					if (typeof coookies.set === 'function') {
```
kangmingtay, line 92: typo
```suggestion
					if (typeof cookies.set === 'function') {
```

kangmingtay (changes requested):
~the package doesn't build with `pnpm build:ssr`~

edit: i've fixed the build errors in [this commit](https://github.com/supabase/auth-helpers/pull/760/commits/d43a5a967bd2d60f9987ebf8aeff53e11d422ced)

packages/ssr/src/createServerClient.ts
```diff
@@ -69,18 +89,22 @@ export function createServerClient<
...
+						// first delete all chunks so that there would be no overlap
+						await deleteAllChunks();
```
kangmingtay: missing the key

packages/ssr/src/utils/chunker.ts
```diff
@@ -99,7 +99,6 @@ export async function deleteChunks(
 
 	if (value) {
 		await removeChunk(key);
-		return;
```
kangmingtay, line 102: removing the return here just makes this function delete all the chunks anyway - we should just remove this part and just keep the for loop below

kangmingtay:
@hf remember to generate a changeset too!

glib-0:
Does this change fix the bug I just submitted? #766

packages/ssr/src/utils/chunker.ts
hf: We need to delete the non-chunked and all potential chunks. Not just remove the non-chunked value.

hf:
> Does this change fix the bug I just submitted? #766

Maybe...

PR #765, opened 2024-04-09
fix: cookie maxAge should be in seconds

hmnd:

Bug fix

Cookie maxAge is currently being set to 1000 years, since maxAge is specified in seconds, not milliseconds.

## What is the new behavior?

Cookie maxAge is 365 days (1 year), below the maxAge limit of 400 days (https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-13#section-4.1.2.2)

### DIFF
--- a/packages/ssr/src/utils/constants.ts
+++ b/packages/ssr/src/utils/constants.ts
@@ -4,5 +4,5 @@ export const DEFAULT_COOKIE_OPTIONS: CookieOptions = {
 	path: '/',
 	sameSite: 'lax',
 	httpOnly: false,
-	maxAge: 60 * 60 * 24 * 365 * 1000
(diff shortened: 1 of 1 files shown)


kangmingtay:
hi @hmnd, this is a non-issue as mentioned in [this comment](https://github.com/supabase/auth-helpers/issues/441#issuecomment-2046645189), will be closing the PR but please reopen it if you think that there's still a need for this

hmnd:
@kangmingtay the reason I brought it up is that it causes my Hono app to break. Hono adheres to RFC 6265 and throws when a cookie with a max age  > 400 days is set.

hmnd:
@kangmingtay I don't think I'm able to reopen, but I think this is still an issue for the reason above :)

PR #769, opened 2024-04-16
feat: auth-helpers-nextjs total fix with createMiddlewareClientV2

hf:
`createMiddlewareClient` was basically completely broken for correctly syncing the session state between the server and browser parts of a NextJS application.

It works in most situations where the user is actively interacting with the site, but as soon as they leave the site for an extended period of time and come back, it's very likely that after the JWT expiry time they will receive an [Invalid Refresh Token](https://supabase.com/docs/guides/auth/sessions#what-is-refresh-token-reuse-detection-and-what-does-it-protect-from) error from the server. Because these events are not time-correlated, it's impossible to debug what is causing them.

Why does this happen:

1. [`NextResponse.next()`](https://nextjs.org/docs/app/api-reference/functions/next-response#next) is not enough. Pages, server-rendered components and route handlers do not see the `Set-Cookie` header set on the *response*.
2. Because this runs in the `middleware.ts` file, the session is correctly refreshed in the middleware client and the `Set-Cookie` headers are set on the response so the browser syncs up with the state *of the middleware client.*
3. However, the page or component client rendered on the server still only sees the `cookie` header from the **request** which is the old session and any use of `getSession()` or `getUser()` will refresh the session *every time.* If you have, say, 3 Data API calls to PostgREST, it's very likely that the session would be refreshed 3 times.
4. Because the page or component cannot set headers on the response, these refreshed sessions just remain unused, but the Auth server already tracks them as the "later" refresh tokens for the user session and is expecting (for security reasons) the browser to know about them too.
5. One hour after this has happened, the Supabase client on browser or server will try to refresh the session from step 2 (coming from `middleware.ts`) and be greeted with the "Invalid Refresh Token" error.

How is this being fixed:

The design of `createMiddlewareClient` does not care at all about what happens to the request, but this is incredibly important so the new refreshed session is passed down to the page or server-rendered component. Furthermore, this cannot "just be set" as for the propagation to take place, `NextResponse.next({ request })` needs to be called.

Therefore, `createMiddlewareClient` is fully deprecated and replaced with its new counterpart `createMiddlewareClient2`. This works a bit differently (without reinventing the simplistic beauty of the `@supabase/ssr` package):

```typescript
export function middleware(request: NextRequest) {
  const [supabaseClient, nextResponse] = createMiddlewareClient2(request, { /* standard client options */ })

  const { data: { user } } = await supabaseClient.auth.getUser()

  if (!user) {
    // there's no user session, ask them to log in
    return NextResponse.redirect(new URL('/sign-in-page-in-your-app'))
  }

  return nextResponse()
}
```

Firstly it returns two arguments, the Supabase client and a `nextResponse` function. It should be called towards the bottom of the `middleware.ts` file, ideally in the `return` statement.

The client is setup such that, it initially reads from the request's `Cookie` header, but writes the changes to local memory. It tracks what **items** (not cookies!) are being set or deleted.

When you call `nextResponse()` the requests's `Cookie` header is reset to reflect the new state of the Supabase client. Because this header will just be propagated down to the page or server-rendered component, no cookie chunking is necessary.

Then, it calls:

```typescript
NextResponse.next({
  request
})
```

Which ensures that the request's changes will be actually propagated down to the page or component. This enables the clients there to pick up the `Cookie` header which now includes the refreshed session.

Finally, it sets the `Set-Cookie` header on the response, which eventually reaches the browser and the session state is finally synchronized between them.

### DIFF
--- a/packages/nextjs/src/middlewareClient.ts
+++ b/packages/nextjs/src/middlewareClient.ts
@@ -15,6 +15,9 @@ import type { NextRequest } from 'next/server';
 import type { GenericSchema } from '@supabase/supabase-js/dist/module/lib/types';
 import type { SupabaseClient } from '@supabase/supabase-js';
 
+/**
+ * @deprecated Broken implementation, do not use.
+ */
 class NextMiddlewareAuthStorageAdapter extends CookieAuthStorageAdapter {
 	constructor(
 		private readonly context: { req: NextRequest; res: NextResponse },
@@ -60,6 +63,11 @@ class NextMiddlewareAuthStorageAdapter extends CookieAuthStorageAdapter {
 	}
 }
 
+/**
+ * @deprecated Use {@link #createMiddlewareClientV2}. This function has a broken
+ * implementation which can cause issues with maintaining the session for a
+ * user for longer periods of time.
+ */
 export function createMiddlewareClient<
 	Database = any,
 	SchemaName extends string & keyof Database = 'public' extends keyof Database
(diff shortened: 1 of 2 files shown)


kangmingtay (dismissed):
`pnpm build:nextjs` fails - also, if `createMiddlewareClient` is broken, why not just modify the implementation to call `createMiddlewareClient2` under the hood instead?

hf:
> `pnpm build:nextjs` fails - also, if `createMiddlewareClient` is broken, why not just modify the implementation to call `createMiddlewareClient2` under the hood instead?

As described in the PR -- to forward the headers `NextResponse.next()` should be called after the client has processed the session.

packages/nextjs/src/middlewareClient.ts
```diff
@@ -102,3 +114,228 @@ export function createMiddlewareClient<
 		}
 	});
 }
+
+class RequestCookiesAuthStorageAdapter extends CookieAuthStorageAdapter {
+	items: { [name: string]: string } = {};
+	deleteItems: { [name: string]: boolean } = {};
```
kangmingtay, line 120: don't really understand why it is important to maintain this extra state here - why can't we just write directly to the cookie everytime `setItem` is called here rather than requiring the developer to call the response fn returned which calls `commitCookies`?

packages/nextjs/src/middlewareClient.ts
```diff
@@ -102,3 +114,228 @@ export function createMiddlewareClient<
...
+ *   if (!user) {
+ *     // there's no user session, ask them to log in
+ *     return NextResponse.redirect(new URL('/sign-in-page-in-your-app'))
+ *   }
+ *
+ *   return nextResponse()
+ * }
+ * ```
+ */
+export function createMiddlewareClient2<
```
kangmingtay, line 241: hmm ok maybe `createMiddlewareClientV2` sounds better?

packages/nextjs/src/middlewareClient.ts
```diff
@@ -102,3 +114,228 @@ export function createMiddlewareClient<
...
+			requestCookies[key] = storage.items[key];
+		});
+
+		// reconstitute the cookie header on the request
+		req.headers.set(
+			'cookie',
+			Object.entries(requestCookies)
+				.map(([key, value]) => `${key}=${value}`)
+				.join('; ')
+		);
```
kangmingtay, line 327: ah so is this where we make sure that the request contains the newly refreshed session so subsequent calls to `getUser` / `getSession` in the server-side components?

packages/nextjs/src/middlewareClient.ts
```diff
@@ -102,3 +114,228 @@ export function createMiddlewareClient<
...
+	Schema extends GenericSchema = Database[SchemaName] extends GenericSchema
+		? Database[SchemaName]
+		: any
+>(
+	req: NextRequest,
+	{
+		supabaseUrl = process.env.NEXT_PUBLIC_SUPABASE_URL,
+		supabaseKey = process.env.NEXT_PUBLIC_SUPABASE_ANON_KEY,
+		options,
+		cookieOptions
```
kangmingtay, line 255: the `cookieOptions` seems unused - i think we need to pass this down or else the custom cookie option won't be supported

packages/nextjs/src/middlewareClient.ts
```diff
@@ -102,3 +114,228 @@ export function createMiddlewareClient<
...
+export function createMiddlewareClient2<
+	Database = any,
+	SchemaName extends string & keyof Database = 'public' extends keyof Database
+		? 'public'
+		: string & keyof Database,
+	Schema extends GenericSchema = Database[SchemaName] extends GenericSchema
+		? Database[SchemaName]
+		: any
+>(
+	req: NextRequest,
```
kangmingtay, line 250: hmm it seems like the only difference between `createMiddlewareClient` and `createMiddlewareClient2`  in terms of the input params is this argument (`createMiddlewareClient` takes in a `context: { req: NextRequest; res: NextResponse }` instead 

which begs the question - why can't we just modify `createMiddlewareClient` to use this new implementation - where we set the cookie on the request after cookie has been updated with the new session?

packages/nextjs/src/middlewareClient.ts
hf: Potentially folks could call getUser twice, which means that the refresh would happen twice. I thought it best to just keep it in memory once set.

About the deletes, deletes are a special "command" on Set-Cookie (set empty with immediate expire) which probably needs to be specifically encoded.

packages/nextjs/src/middlewareClient.ts
hf: I think it's doable, but I want to signal to users that they need to move to a different way of use. A new method name seems to signal this more effectively than signature/type change, though that's also an option.

We need to keep both ways to prevent complete compile-time breakage (even though functionality is broken) to not disrupt folks unnecessarily.

kangmingtay: or we could change the functionality, bump the major version of the package and indicate in the changelog that the underlying implementation of `createMiddlewareClient` is different

hf: I don't want to bump the major version because nobody will upgrade to that, due to `^`. The old method remains just so they are able to compile, it will show up crossed-out in VScode and will add a postinstall log message.

packages/nextjs/src/middlewareClient.ts
kangmingtay: but is there any downside to just writing to the cookie directly? the cookie is also stored in memory so i'm not sure what the actual benefit is here?

packages/nextjs/src/middlewareClient.ts
```diff
@@ -82,6 +90,10 @@ export function createMiddlewareClient<
 		cookieOptions?: CookieOptionsWithName;
 	} = {}
 ): SupabaseClient<Database, SchemaName, Schema> {
+	console.warn(
```
kangmingtay, line 93: maybe `console.debug` instead? not sure if we want to repeat the mistake of flooding the console with a warning here

kangmingtay:
@hf you need to export the new client in the `index.ts` file too

J0:
Reviewing now

packages/nextjs/src/middlewareClient.ts
hf: Removed it.

kangmingtay (approved):

mandarini:
I'll be closing this, since the move to [@supabase/ssr](https://github.com/supabase/ssr). Let me know if I should reopen!

PR #771, opened 2024-04-19
chore(ssr): replace ramda with moderndash, reduce global package size

Olyno:

The SSR package does not have a lot of dependencies, but one of them is taking the whole place. I tried multiple tools to compare the size, and it looks ``ramda`` is the one having this issue.

I did some search, and the lightest package i found (not deprecated) was [``moderndash``](https://github.com/Maggi64/moderndash), offering around -75% less in terms of size, while still having the same features.

Some requirements are needed tho:
- Node version >= 16.5
- Typescript (if used) => 4.8

The package is too heavy for the code inside it.

## What is the new behavior?

This change reduces the size of the package. Not sure of how many, but it should reduce the size of 50% at least.

### Bundlephobia

image

### pkg-size.dev

image

### Npmjs

image

### DIFF
--- a/packages/ssr/package.json
+++ b/packages/ssr/package.json
@@ -38,8 +38,7 @@
 		"node": ">=16.15"
 	},
 	"dependencies": {
-		"cookie": "^0.5.0",
-		"moderndash": "^3.11.1"
+		"cookie": "^0.5.0"
 	},
 	"devDependencies": {
 		"@supabase/supabase-js": "2.42.0",
--- a/packages/ssr/package.json
+++ b/packages/ssr/package.json
@@ -34,19 +34,22 @@
 		"url": "https://github.com/supabase/auth-helpers/issues"
 	},
 	"homepage": "https://github.com/supabase/auth-helpers#readme",
+	"engines": {
+		"node": ">=16.15"
+	},
 	"dependencies": {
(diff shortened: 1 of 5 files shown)


hf:
Actually this whole library is unnecessary only for merge. If you want you can remove it with a local implementation of merge.

silentworks:
I just want to add here that at the time of working on this only `ramda` had support for `mergeRight`, note this isn't just a normal shallow merge or normal deep merge. Please take a look at what `ramda` mergeRight does before removing it.

Olyno:
From what i'm reading

> Create a new object with the own properties of the first object merged with the own properties of the second object. If a key exists in both objects, the value from the second object will be used.

From my understanding, it is the same thing than a simple a shallow or deep merge, as it combines both objects by replacing values, and returns a new object. Except if i'm wrong, [the code i'm using](https://github.com/supabase/auth-helpers/pull/771/files#diff-3e067c41e610e72de0a58a35fa26c2a3f3123eb858668462d3c81fcffcacbb49R7-R20) should do the job 🤔 

Please tell me your thought about it @silentworks

hf:
This is getting "fixed" in this PR: https://github.com/supabase/ssr/pull/1

PR #776, opened 2024-04-28
fix: cookie maxAge should be in seconds

hmnd:

Bug fix

Cookie maxAge is currently being set to 1000 years, since maxAge is specified in seconds, not milliseconds.

## What is the new behavior?

Cookie maxAge is 365 days (1 year), below the maxAge limit of 400 days (https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-13#section-4.1.2.2)

## Additional context

Sorry for creating #765 again, but I don't think it's being monitored anymore now that it is closed. 

As detailed in the earlier PR, this is seriously affecting usability for us with Hono, as it's currently requiring me to re-patch `@supabase/ssr` every time a new update to the package is released.

### DIFF
--- a/packages/ssr/src/utils/constants.ts
+++ b/packages/ssr/src/utils/constants.ts
@@ -4,5 +4,5 @@ export const DEFAULT_COOKIE_OPTIONS: CookieOptions = {
 	path: '/',
 	sameSite: 'lax',
 	httpOnly: false,
-	maxAge: 60 * 60 * 24 * 365 * 1000
+	maxAge: 60 * 60 * 24 * 365
 };


j4w8n:
I'm pro rfc-adherence. The only pushback I can possibly see is that the Proposed Standard rfc 6265 does not define a limit; only this new draft does.

It may have been unwise for Hono to have a hard failure for this; but at the same time, 365 days, or 400, seems reasonable and easily changeable by Supabase.

If anyone logs into a website, then closes the browser and waits a year to go back to that site, it's reasonable to expect them to log in again.

hmnd:
@j4w8n fyi, Chrome, Safari, and Firefox have all agreed on this limit and at least Chrome has shipped this limit in v104. Nonetheless, I do agree that Hono shouldn't be forcing this onto its users and have thus opened an issue there too.

mandarini:
Thank you for your PR and for contributing to Supabase! I will be closing this PR since we have moved the functionality of `auth-helpers` into the [@supabase/ssr](https://github.com/supabase/ssr) package. Please direct your contributions there. Thank you for understanding and for being a member of our community!

PR #781, opened 2024-05-17
move @types/cookie to main dependencies 

rbutera:

Bug fix / Dependencies update

Currently `@types/cookie` is listed as a dev dependency only which results in projects that use `@supabase/ssr` running into difficult-to-debug `@typescript-eslint` `no-unsafe-argument` errors. See [supabase/auth issue supabase/ssr#53](https://github.com/supabase/ssr/issues/53) where this behaviour is reported and documented.

The change in this PR satisfies the solution proposed in the above PR.

## What is the new behavior?

`@types/cookie` is added to dependencies so that the `CookieOptions` type can be properly inferred.


mandarini:
Thank you for your PR and for contributing to Supabase ! I will be closing this PR since we have moved the functionality of `auth-helpers` into the [@supabase/ssr](https://github.com/supabase/ssr) package. Please direct your contributions there. Thank you for understanding and for being a member of our community!

